Issue #105, opened 2014-05-23
"Delete a job if its latency is too high" filter


odinserj:
This filter should be implemented as `IElectState` filter that processes jobs that are going to move to the `ProcessingState` and change the state to the `DeletedState`, if a latency (current time - creation time) exceeds a given time span.

```csharp
[DeleteOnLatencyTimeout(timeoutInSeconds: 5 * 60)]
public void SampleMethod()
{
}
```

odinserj:
Closing the issue, #188 was merged.

PR #188, opened 2014-08-28
Add Latency Filter Attribute To Delete job if timeout is exceeded

Excommunicated:
This is for Issue #105 
Added attribute based on the discussion in the listed issue..
Please let me know if there is anything I have misinterpreted.


llamaluvr:
Are there any plans to merge in this change? This sounds really useful. Or is there another way to fail a job after a specified timespan has been exceeded?

odinserj:
Okay, this filter was finally merged into Hangfire under the [`LatencyTimeoutAttribute`](https://github.com/HangfireIO/Hangfire/blob/612b44d93c11a1834d3dd859fe489d27cf01c4fd/src/Hangfire.Core/LatencyTimeoutAttribute.cs) name. It really acts on the basis of latency, and applied to a background job that is going to be processed right now (and this is not the processing time limit), so the naming is fine.

@Excommunicated, thank you for this, and for patience :+1: :heart: I've merged it using git command line, preserving your authorship (03a34145539de37fb26a9cab4e96f48c3c860eef), so I'm closing this PR instead of merging it.

PR #199, opened 2014-09-03
Display missing method information on the Dashboard

### LINKED ISSUE #157: Show missing method and its type in the Dashboard
Currently Hangfire shows only unmeaningful _Can not find the target method_ error message in case when the method referred by a job was removed from the code base. It is hard to debug these issues, so the message should be replaced with a more meaningful one:

_Can not find the `Namespace.Type.MethodName(Type1, Type2)` method._.

sgrassie:
Starts to resolve #157. It is marked as 'Jump in', so here goes...

`InvocationData` almost gives us what we need when it throws an `InvalidOperationException` on deserialising if the method is null. The actual exception message thrown is almost exactly what should be shown in the Dashboard.

My idea with this PR will be to stop `Invocation.Deserialize` from throwing the `InvalidOperationException` if the target method is null, and instead add a couple of properties to the job: a boolean property which will indicate the target method is actually missing, and a text property that will contain the text of the `InvalidOperationException`.

Then in `JobMethodCallRenderer.Render` instead of

```
if (job == null) { return new NonEscapedString("<em>Can not find the target method.</em>"); }
```

it could be replaced with

```
if (job.MethodMissing) { return new NonEscapedString(String.Format("<em>{0}</em>"), job.MissingMethodError); }
```
# TODO
- [x] Add `bool methodMissing` to internal `Job` ctor, and associated public property
- [x] Fix tests
- [x] ~~Change `JobFacts.Ctor_ThrowsAnException_WhenTheTypeDoesNotContainTheGivenMethod` to be assert that the ctor will not in fact thrown an exception.~~
- [x] ~~Add additional unit tests to `JobFacts` to assert that the missing method error information is as expected.~~
- [x] Update tests in `InvocationDataFacts` to verify expected behaviour.
- [x] Add `MissingMethodError` property to `Job`
- [x] Change the behaviour in `Invocation.Deserialize` to populate the `MissingMethodError`.
- [x] Change `JobMethodCallRenderer.Render` as detailed above.

### DIFF
--- a/src/Hangfire.Core/Common/Job.cs
+++ b/src/Hangfire.Core/Common/Job.cs
@@ -44,7 +44,7 @@ namespace Hangfire.Common
         /// 
         /// <exception cref="ArgumentNullException"><paramref name="method"/> argument is null.</exception>
         /// <exception cref="ArgumentNullException"><paramref name="arguments"/> argument is null.</exception>
-        internal Job(Type type, MethodInfo method, string[] arguments, bool methodMissing)
+        internal Job(Type type, MethodInfo method, string[] arguments)
         {
             if (type == null) throw new ArgumentNullException("type");
             if (method == null) throw new ArgumentNullException("method");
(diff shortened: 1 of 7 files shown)


OldBayFries:
Please forgive a n00b, but has this been resolved? I still get _Can not find the target method_ in the dashboard for some jobs that ran to completion and are marked Succeeded. The Job database has the invocation data & arguments as expected. Not sure where the disconnect occurs.

gandarez:
@OldBayFries you might copy all those libraries to bin folder across all Hangfires servers/dashboard.

PR #204, opened 2014-09-05
Add a job activation context.

lowkay:
To enable IoC containers to create per-job scopes they need to be able
to know when the job has finished so that the container can be
reclaimed.

A ``JobActivationContext`` exposes an event that can be used by a
``JobActivator`` to clean up a job-scoped container.

### DIFF
--- a/src/Hangfire.Core/Common/Job.cs
+++ b/src/Hangfire.Core/Common/Job.cs
@@ -75,13 +75,16 @@ namespace Hangfire.Common
             object result = null;
             try
             {
-                if (!Method.IsStatic)
+                using (var jobActivationContext = new JobActivationContext())
                 {
-                    instance = Activate(activator);
-                }
+                    if (!Method.IsStatic)
+                    {
+                        instance = Activate(activator, jobActivationContext);
+                    }
 
-                var deserializedArguments = DeserializeArguments(cancellationToken);
-                result = InvokeMethod(instance, deserializedArguments);
+                    var deserializedArguments = DeserializeArguments(cancellationToken);
+                    result = InvokeMethod(instance, deserializedArguments);
+                }
             }
             finally
             {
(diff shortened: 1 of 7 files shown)


kieranbenton:
Is this on the cards to be integrated into the project? I have pretty much exactly the same use-case as @lowkay and my current solution is not clean at all

Dresel:
Is is safe for now to use a job filter to determine the job context (see [SO](http://stackoverflow.com/questions/27961210/hangfire-dependency-injection-lifetime-scope/28034217#28034217))?

sjuranek:
I'm looking at basically this same scenario, where I need 'state' from the time the method was enqueued to Hangfire so we can re-create the state for our IoC container (StructureMap). In our case the IoC will create the DbContext but needs to know the current user. The way hangfire proposes to do this is to put the state into the method as parameters or in the database. That is great but those parameters aren't present in the JobActivator where you use your IoC to create the type. I've fork'd the source and would like to propose simply adding the Job as a parameter to the Activate method:
```
    public class JobActivator
    {
           public virtual object ActivateJob(Type jobType, Job job)
    }
```
Once you have a Job reference you have access to the Method and Arguments where I'd grab the state from. I'm a newbie at pull requests or even if the project owner would want this request so just writing to see if people agree and then I'll get it submitted. We need this functionality asap.

odinserj:
Hi there! In Hangfire 1.5.0-beta1 I've added the JobActivator.BeginScope method as well as the base JobActivatorScope class to allow support for custom scopes/child containers without any job filters to simplify the user experience. Please see my answer on a corresponding forum topic for some details.

Will this new feature close this PR?

lowkay:
Howdy @odinserj yes this new behavior does appear to be able to solve the problem for us. We will try out the latest release to see if we get the expected behavior. Thanks!

odinserj:
@lowkay, ok, so I'm closing this PR. Please feel free to tell more details.

PR #213, opened 2014-09-16
Include write isolation level in SqlStorageOptions

oeaoaueaa:
Hi, we had a problem with HangFire defaulting isolation level to Serialized, so I have made some changes to HangFire.SqlServer so the write job isolation level can be configured as part of SqlServerStorageOptions

original problem: http://discuss.hangfire.io/t/hangfire-sqlserver-defaults-write-isolation-level-to-serializable/224

Let me know what you think.

Thank you

### DIFF
--- a/src/Hangfire.SqlServer/SqlServerConnection.cs
+++ b/src/Hangfire.SqlServer/SqlServerConnection.cs
@@ -32,16 +32,19 @@ namespace Hangfire.SqlServer
     {
         private readonly SqlConnection _connection;
         private readonly PersistentJobQueueProviderCollection _queueProviders;
+        private readonly IsolationLevel _writeIsolationLevel;
 
         public SqlServerConnection(
             SqlConnection connection, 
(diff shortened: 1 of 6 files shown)


jpero:
We are having an issue with Hangfire's use of IsolationLevel.Serializable.  It appears that due to connection pooling, our application's DAL code is occasionally running SQL queries under the SERIALIZABLE isolation level.

This PR was closed a few years back due implementation concerns.  

Looking at the current SqlServerStorageOptions, it appears that you _can now_ configure Hangfire to use a specific isolation level:

`GlobalConfiguration.Configuration
	.UseSqlServerStorage("Server=.;Database=HangfireDB;Trusted_Connection=True",
		new SqlServerStorageOptions { TransactionIsolationLevel = System.Transactions.IsolationLevel.ReadCommitted })
	);`

Are there existing concerns we should know about, when running Hangfire using READ COMMITTED iso level?

PR #220, opened 2014-09-27
.NET 4.0 support

### LINKED ISSUE #207: .net 4 support?
RT

phillcampbell:
This PR is an update to #216  and fixes #207.

Support for .NET 4.0 is now added through dedicated project files, rather than using conditionals. Additionally .NET 4.0 tests are built and run in the build scripts.

Unfortunately I have had to add this as another PR, as GitHub doesnt support changing the branch of the original PR. 

- [ ] AppVeyor build and tests passing
- [x] Mono build and tests passing

### DIFF
--- a/Hangfire.Mono.net40.sln
+++ b/Hangfire.Mono.net40.sln
@@ -1,23 +1,25 @@
 ﻿
 Microsoft Visual Studio Solution File, Format Version 12.00
-# Visual Studio 2012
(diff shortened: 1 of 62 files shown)


odinserj:
Won't support .NET 4.0 as it greatly complicates everything, including project structure, build tools, packaging, etc. I think there will be enough challenges with the upcoming .NET Core.

Issue #218, opened 2014-09-25
Log custom messages to the job history page?


ses4j:
The job history page currently shows Processing, Succeeded, etc.  Is it possible for a job to write arbitrary logging (for instance, periodic "current status" messages or warnings) into it? If not, is that a feature you are interested in?

samcalma:
How can we help on getting this implemented?

vip32:
+1

efirdman:
+1

pascalberger:
+1

peters:
+1

searope:
+1

oliverkane:
+1

ses4j:
Note that pull request #232 (referenced above but maybe lost in the chain) implements this.

RuiRomano:
+1

dazinator:
+1

samshiles:
Please see: https://github.com/samshiles/hangfire.job.logging for one way to get job output logged to the hangfire console. It requires a very small change to hangfire which should be 100% backwards compatible. PR: #466

IanGrainger:
+1

stajs:
+1

vip32:
@samshiles : first thanks for sharing the idea to insert in the state table directly.
however, you don't need to inject the jobid, using a jobcontext makes it much easier and transparant
this can be used inside your joblogger to get the right jobid. i just did that in my project and works really well.

public class JobContext : IServerFilter
    {
        [ThreadStatic]
        private static string _jobId;
        [ThreadStatic]
        private static Type _jobType;

        public static string JobId { get { return _jobId; } set { _jobId = value; } }

        public static Type JobType { get { return _jobType; } set { _jobType = value; } }

        public void OnPerforming(PerformingContext context)
        {
            JobId = context.JobId;
            JobType = context.Job.Type;
        }

        public void OnPerformed(PerformedContext filterContext)
        {
        }
    }

jasoncouture:
+1

DonWMason:
+1

cocowalla:
+1

gandarez:
+1

Mykezero:
+1

MaxMelcher:
+1

mrcrbrown:
+1

migajek:
+1

sergeyzwezdin:
Any update on this?

smith288:
+1

dandcg:
+1

[...]

PR #232, opened 2014-10-23
Added logging support to Hangfire

ses4j:
Covers issues raised in issue #218, by augmenting the cancellation token with a more feature-ful JobExecutionContext (which inherits the cancellationtoken for backwards compatibility reasons).  

JobExecutionContext offers an API for logging and accessing the Job Id.  Logs are written to the existing state table as "PROCESSING" messages.  Possibly, some table changes should be made to organize things a little cleaner, but I didn't know how to deal with migration issues, so I worked with the existing schema.

The dashboard displays logged messages as seen below.

capture


ses4j:
@odinserj I just rebased this pull request to the current master.  Is there any chance of it being merged in in some form?

mnesler:
Hi ses4j, funny we were looking for this just over a week ago...is there an example somewhere with how to use this?  Thanks, max

pascalberger:
+1

ses4j:
@mnesler The simplest way to use it is:

    JobExecutionContext.Current.LogDebug("Hello world!");

You can try any of these: `context.LogDebug()`, `context.LogInfo()`, `context.LogWarn()`, `context.LogError()`, `context.LogFatal()`.  `context` equally comes from the `JobExecutionContext.Current` singleton or from accepting the JobExecutionContext token the same way as the mainline branch accepts the CancellationToken.

yngndrw:
@ses4j I'd suggest that ...
```cs
JobExecutionContext.Current = null;
```
... is put inside a finally block.

I'm not sure if JobExecutionContext.Current should be set or unset for the SucceededState constructor, but I think it's correct that it is not set and therefore think that it should happen outside of the finally block.

mnesler:
@ses4j thank you for the response, sorry i got back so late!

src/Hangfire.Core/Dashboard/Pages/JobDetailsPage.cshtml
```diff
@@ -119,11 +119,17 @@
...
                 var index = 0;
 
-                DateTime[] entriesCreationTime = job.History.Select(x => x.CreatedAt).ToArray();
+                var stateGroups = job.History.ChunkBy(x => x.StateName);
+        
+                DateTime[] entriesCreationTime = stateGroups.Select(x => x.Last().CreatedAt).ToArray();
                 var nextEntry = 1;
 
-                foreach (var entry in job.History)
+                foreach (IGrouping<string, Hangfire.Storage.Monitoring.StateHistoryDto> stateGroup in stateGroups)
```
nirinchev: Can this be a ```var``` or at least get rid of the namespace?

src/Hangfire.Core/Dashboard/Pages/JobDetailsPage.cshtml
```diff
@@ -119,11 +119,17 @@
...
                 {
+                    var entry = stateGroup.Last();
+                    var mostRecentEntry = stateGroup.First();
+                    
+                    var stateEnteredAt = entry.CreatedAt;
```
nirinchev: Where is this used?

src/Hangfire.Core/Dashboard/Pages/JobDetailsPage.cshtml
```diff
@@ -119,11 +119,17 @@
...
+                var stateGroups = job.History.ChunkBy(x => x.StateName);
+        
+                DateTime[] entriesCreationTime = stateGroups.Select(x => x.Last().CreatedAt).ToArray();
                 var nextEntry = 1;
 
-                foreach (var entry in job.History)
+                foreach (IGrouping<string, Hangfire.Storage.Monitoring.StateHistoryDto> stateGroup in stateGroups)
                 {
+                    var entry = stateGroup.Last();
+                    var mostRecentEntry = stateGroup.First();
```
nirinchev: This seems unused, or am I missing something?

src/Hangfire.Core/IJobExecutionContext.cs
```diff
@@ -0,0 +1,37 @@
...
+
+        /// <summary>
+        /// Report the current status of a job, potentially to an end-user, along with an
+        /// estimate of the % of the job that is completed.
+        /// </summary>
+        /// <param name="percentComplete"></param>
+        /// <param name="currentStatus"></param>
+        void UpdateProgress(int percentComplete, string currentStatus);
+        
+        void LogDebug(string message);
```
nirinchev: Is there a real need to have 5 methods for different severity levels instead of a single ```Log(Severity severity, string message```? Where ```Severity``` is an enum containing the different levels.

src/Hangfire.Core/States/ProcessingLogState.cs
```diff
@@ -0,0 +1,32 @@
...
+    public class ProcessingLogState : IState
+    {
+        public ProcessingLogState(string level, string message)
+        {
+            this.Level = level;
+            this.Reason = message;
+        }
+
+        public string Name { get { return ProcessingState.StateName; } }
+        public string Reason { get; set; }
```
nirinchev: Any reason for a public setter to be exposed? Same goes for Level.

src/Hangfire.Core/States/ProcessingLogState.cs
```diff
@@ -0,0 +1,32 @@
+﻿using System;
+using System.Collections.Generic;
+using System.Linq;
+using System.Text;
+using System.Threading.Tasks;
+
+namespace Hangfire.States
+{
+    public class ProcessingLogState : IState
```
nirinchev: This looks a lot like ```ProcessingSubState```. Any chance to extract an abstract base?

tests/Hangfire.Core.Tests/Common/JobArgumentFacts.cs
```diff
@@ -13,7 +13,7 @@ namespace Hangfire.Core.Tests.Common
 	public class JobArgumentFacts
 	{
 		private readonly Mock<JobActivator> _activator;
-		private readonly Mock<IJobCancellationToken> _token;
+        private readonly Mock<IJobExecutionContext> _jobExecutionContext;
```
nirinchev: Whitespace issues here - it seems the project is using tabs and you're using spaces, so probably you'd like to tabify the 5 changed files.

src/Hangfire.Core/IJobExecutionContext.cs
ses4j: I think it's pretty standard practice, and useful.  The ILog/LogExtensions hangfire uses internally does the same thing..

tests/Hangfire.Core.Tests/Common/JobArgumentFacts.cs
ses4j: will fix

src/Hangfire.Core/Dashboard/Pages/JobDetailsPage.cshtml
ses4j: I'll remove.  I was playing with showing the full time range of the action.

src/Hangfire.Core/Dashboard/Pages/JobDetailsPage.cshtml
ses4j: I did that because VS (not csc) is confused by the type coming out of ChunkBy, and to get intellisense working for the `stategroup` I told it the right type.

yngndrw: You could use an alias instead.

pascalberger: You should be able to do a Cast<StateHistoryDto>() on stateGroups to avoid the type

ses4j: Ya I could do those things.  Why is the explicit type so frowned upon, though?

src/Hangfire.Core/Dashboard/Pages/JobDetailsPage.cshtml
ses4j: Entry is used, mostRecentEntry not used. Removing.

src/Hangfire.Core/States/ProcessingLogState.cs
ses4j: I don't think it's worth it at this point as these things are almost structs, but one could...

src/Hangfire.Core/Dashboard/Pages/JobDetailsPage.cshtml
nirinchev: I don't mind the explicit type (actually using .Cast will make it just as long). It's just that the fully qualified class name seems a bit verbose. So if you need to specify the type for intellisense to kick in, either use an alias, or at least shorten it to ```IGrouping<string, StateHistoryDto>``` :)

src/Hangfire.Core/IJobExecutionContext.cs
nirinchev: I too think it's useful, but specifying it in the interface, forces every implementor to copy past much of the same implementation, which pollutes the contract. A better approach (in my opinion) would be to have only ```Log(Severity severity, string message)``` in the interface, and then implement the helper ```LogDebug```, ```LogInfo``` and so on as extension methods. This keeps the contract clean, while still preserving the useful helper methods.

ses4j:
Fyi I don't believe the appveyor check failure is legitimate, does anyone know?

strofimo:
Hi ses4j,

Great job! BTW do u have any ideas when this feature will be in official package?

thanks

lucashby:
+1

Will this be in an upcoming release?

Balrog30:
If this works, I like the sounds of what it does. I was looking for exactly this. I handle some Exceptions in code that runs in the Hangfire context. I was the code to fail gracefully and continue (it's fetching some optional data from a third-party API), but I want to be able to check the logs and see if there was a silent failure.

I think this might technically count as a breaking change though, since this changes the name of an existing interface in the public API.

Elph:
+1

samshiles:
+1, would love to see this merged in. Any plans?

dazinator:
+1

samshiles:
I actually tried to merge this into the current master today (locally). Unfortunately the implementation on this PR targets a number of the now obsolete method of Job.cs. It's by no means impossible to redo it using the new interface, but it might require a little rethought.

samshiles:
Please see: https://github.com/samshiles/hangfire.job.logging for one way to get job output logged to the hangfire console. It requires a very small change to hangfire which should be 100% backwards compatible. PR: https://github.com/HangfireIO/Hangfire/pull/466

odinserj:
Hey guys, see what @pieceofsummer did in the [Hangfire.Console](https://github.com/pieceofsummer/Hangfire.Console) extension, even with colours and live updates! Almost all storages are supporting this with no changes!

dashboard

PR #262, opened 2014-11-30
Allow to customize serialization settings of Json.NET

fpellet:
I don't like use global configuration of application, but a specific configuration to facilitate maintenance


vp123456:
Is there a work around for this? Beside setting global settings?

Is there a target date for Hangfire 2?

PR #263, opened 2014-11-30
Inject a factory to customize component creation

fpellet:
There are lot of static methods in this library
Thank's to use less statics and inject service. It's better to evolve and customize.
Open Close principe of SOLID.
It's my first pull request to improve this problem


odinserj:
I hate the current implementation of the `BackgroundJobServer` class, but I've found a way to decomposite it to have an ability to pass custom collections of the `IServerComponent` interface implementations. This will decrease the coupling for the class and will allow to use all the features implemented by your other PRs:

* Inject a factory to customize component creation (#263).
* Allow to custom Filters for each server instance (#374).
* Allow to custom JobActivator for each server instance (#373).

That way is clean and elegant, however requires a lot of breaking changes that may be introduced only in version `2.0` :frowning:. However, all of these features can be partly implemented in version `1.x`. So let's do them in the following way as all of them add a new constructor overload and thus there will be merge conflicts.

I'll close this PR and add support for putting additional custom `IServerComponent` implementations into `BackgroundJobServer` instance. Then I'll ping you in #374 to merge the changes, and after merging #374 we'll merge #373. What do you think?

fpellet:
:+1:  
I like a IServerComponent collection :) 
You can use also builder pattern. It's more extensible with other libs, as owin extension method.

odinserj:
@fpellet, please see https://github.com/HangfireIO/Hangfire/pull/374#issuecomment-124475247 and https://github.com/HangfireIO/Hangfire/pull/374#issuecomment-124481327. There were some breaking changes, however these changes concert users of a low-level API, like you. I think it is better to allow them to use more functionality, then wait for 2.0.

PR #297, opened 2015-01-31
Fix GetTimelineStats / GetHourlyTimelineStats incorrect order

Dresel:
Hi,

`GetTimelineStats` and `GetHourlyTimelineStats` job counts may be assigned to the wrong dates. The problem is that in the last loop where the result is generated the matching of the value of `valuesMap[valuesMap.Keys.ElementAt(i)]` and the date `dates[i]` are not correct.

    for (var i = 0; i < dates.Count; i++)
    {
        var value = valuesMap[valuesMap.Keys.ElementAt(i)]; // valuesMap.Keys may be unsorted if keys were missing and added via valuesMap.Add(key, 0);
        result.Add(dates[i], value); // dates are sorted
    }

I also refactored some duplicated code into a separate function.

### DIFF
--- a/src/Hangfire.SqlServer/SqlServerMonitoringApi.cs
+++ b/src/Hangfire.SqlServer/SqlServerMonitoringApi.cs
@@ -369,31 +369,9 @@ select count(*) from HangFire.[Set] where [Key] = N'recurring-jobs';
                 endDate = endDate.AddHours(-1);
             }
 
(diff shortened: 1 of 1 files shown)


odinserj:
Love you! :-D Planned to 1.3.4 coming soon.

odinserj:
And thank you for switch from `dev` to `master`, this will help to release it sooner.

Dresel:
I have to thank (love) you for this project - keep on :)

PR #302, opened 2015-02-04
Making Product Version and File Version on the dll match the Assembly Version

fire-eagle:
Having the AssemblyFileVersion attribute defined causes the product version and file version in the dll properties to follow it. Removing it makes those properties follow the assembly version. This was causing issues within a MSI my project uses, since the MSI thinks the version is always 1.0.0.0 and skips deploying the new dll.

### DIFF
--- a/src/SharedAssemblyInfo.cs
+++ b/src/SharedAssemblyInfo.cs
@@ -12,4 +12,3 @@ using System.Runtime.InteropServices;
 
 // Don't edit manually! Use `build.bat version` command instead!
 [assembly: AssemblyVersion("1.4.0")]
-[assembly: AssemblyFileVersion("1.0.0.0")]


odinserj:
Do you use Newtonsoft.Json? It also has a frozed version for the assembly. Maybe there are some ways to tell MSI to use modification date or so?

fire-eagle:
Which version of Newtonsoft.Json.dll has that issue? The projects I'm responsible for use v6.0.8, and the version attributes in the dll properties appear to be filled out correctly in that version.

There are ways to make the MSI install an older version, but they involve either deleting the old dll with an additional action or turning off some behaviour across the entire msi. Unless this causes a problem on the Hangfire side, I would rather just get the attributes updating properly so the MSI can use it's normal behaviour around upgrading.

odinserj:
Hm you are right, Json.NET [uses fixed](https://github.com/JamesNK/Newtonsoft.Json/blob/eb1c683de389b5b08bd1a37d345d9c8f54b4de6b/Src/Newtonsoft.Json/Properties/AssemblyInfo.cs#L93) `AssemblyVersion` values and dynamic `AssemblyFileVersion` values. So, merging. Thanks for the tip.

PR #303, opened 2015-02-06
Allow a URI to be used for RabbitMQConnectionConfiguration

dennyferra:
This change allows a URI to be used for the RabbitMQConnectionConfiguration. Using an AMQP URI is less verbose than having to specify each configuration property. This is backwards compatible, if no URI is specified then the default properties are expected to be set.


odinserj:
@dennyferra, sorry for the delay. Can you re-target this PR to the `dev` branch?

dennyferra:
Created new PR https://github.com/HangfireIO/Hangfire/pull/321

PR #309, opened 2015-02-23
Revert 2 revert 1 sjuranek patch 1

sjuranek:


sjuranek:
Sorry, my first time with a pull request so was just trying a simple update to the readme.

nirinchev:
Since we can safely assume no one's based off your branch, you can squash all your commits and push force it so that the history looks neat.

sjuranek:
I ended up just deleting my fork and starting over. I since learned I also need to create a branch for each feature (or ideally). Is there a way to just delete this pull request?

nirinchev:
I think you can simply close it. As far as I know, github doesn't allow deletion.

PR #310, opened 2015-02-23
Added prerequisites for building the solution with the build script. Also modified JobActivator to accept the Job parameter so activator has a chance to get some of the original context back before creating an instance.

sjuranek:


sjuranek:
Not necessary as I was introduced to JobFilters by Sergey.

PR #311, opened 2015-02-23
Expose deserialized job arguments

sjuranek:
The Job object now exposed the de-serialized arguments rather than just the string array of arguments. Makes it easier for person implementing the JobActivator to get back to the original intent of things.


sjuranek:
Not necessary as I was introduced to JobFilters by Sergey.

odinserj:
Thanks for closing. Will provide more examples of extensibility after releasing the next version.

PR #321, opened 2015-02-28
Allow a URI to be used for RabbitMQConnectionConfiguration

dennyferra:
This change allows a URI to be used for the RabbitMQConnectionConfiguration. Using an AMQP URI is less verbose than having to specify each configuration property. This is backwards compatible, if no URI is specified then the default properties are expected to be set.

* Updated to point to HangFire Dev branch


odinserj:
Thanks, Danny, merging.

PR #323, opened 2015-03-03
Monthly cron expression that runs on the first specified day in the month

dennisroche:
Monthly cron expression that runs on the first specified day in the month, e.g. the first Tuesday of April.

Using the Crontab syntax, the `Cron.Monthly(DayOfTheWeek.Tuesday, 9)` will return `0 9 1-7 * 2`. This will trigger at 9am on the first Tuesday of any month, with the 1-7 matching any day in the first week.

I'm not certain the expression is correct (I'm not a cron expert) and I'm going to add some further testing to confirm.


tests/Hangfire.Core.Tests/CronScheduleFacts.cs
```diff
@@ -0,0 +1,39 @@
...
+            CrontabSchedule schedule = CrontabSchedule.Parse(expression);
+
+            DateTime expected = new DateTime(2015, 02, 07, 9, 6, 0);
+            DateTime actual = schedule.GetNextOccurrence(baseTime);
+
+            Assert.Equal(expected, actual);
+        }
+
+        [Fact]
+        public void Monthly_WithDayOfTheWeek_ThereIsOnlyOccuranceInNextMonthAndItIsTheFirstTuesdayOfThatMonth()
```
dennisroche: These are the new tests that I would like to add to check the cron expressions. What do you think?

If you like, then I will add a test for each of the expressions we use.

dennisroche:
Hi @odinserj, do you need anything further to merge this PR?

Issue #270, opened 2014-12-04
Allow Recurring Tasks to be Scheduled for Local Time


jkodroff:
The changes made for #160 don't work well for my use case:

My client is a smaller organization (i.e. a single time zone) which needs a batch of emails sent every night at 7:00 PM (1900h) (after their employees leave for the day) so it *should* respect daylight savings time.

Here's my workaround for the current UTC-only limitation:
```
            RecurringJob.AddOrUpdate(
                "upcoming-jobs-email", 
                () => ObjectFactory.GetInstance<UpcomingJobsEmails>().Run(),
                "0 {0} * * 1-5".ToFormat(DateTime.Today.AddHours(19).ToUniversalTime().Hour)
            );
```

But I believe once my time zone switches in or out of DST, I'll be off by an hour until the app pool recycles (which doesn't happen normally for my server setup).

Ultimately, I'd love for there to be an option to schedule tasks for *both* UTC and local time.

jkodroff:
Any word on this?  The job needs to fire at 7:00 PM, and I'm GMT -5, so that means I have to offset the day of the week in addition to the time.  It's really a bummer.

cneller:
Agreed - we would love to see local time threaded throughout.

jkodroff:
I think it makes sense to have either option because in my use case, it's part of the business rules: "At 7:00 every night (Eastern Time) send this report...."

But not everyone's rules will be like that.

mexner:
Agreed! Would love to see the ability to specify a TZ for recurring task.

odinserj:
Time zones is an essential part of recurring jobs, I agree. It was a strange decision to use UTC time zone by default instead of a local one, as all other frameworks and programs do that for CRON-based launchers. Looks like I was drunk ;)

I have a [simple implementation](https://github.com/HangfireIO/Hangfire/tree/recurring-timezones) for this, now thinking about time zone identifiers on Windows and Linux machines. Since ASP.NET vNext makes great strides to support Linux as well, the fact that time zone identifiers are different on different platform should be considered. So, there are the following ways:

* Use BCL time zones (`TimeZoneInfo` class) and solve issues when they are arising.
* Use TZDB as a time zone provider. This may be accomplished with using `NodaTime`.

What do you think?

![image](https://cloud.githubusercontent.com/assets/1078718/6671313/805574a2-cc15-11e4-8869-392ff04da031.png)

mexner:
Love this! Thumbs up.

[...]

Issue #160, opened 2014-07-09
RecurringJobScheduler should use UTC time


sjwoodard:
Everywhere that _dateTimeProvider.GetNextOccurrence(cronSchedule) is used in RecurringJobScheduler should be appended with .ToUniversalTime(). That way if the server and client are in different time zones, the scheduling will make sense.

This creates a problem of converting CRON strings to UTC. I'm experimenting with using the system's UTC offset to move the CRON hours value around. *(EDIT: see next post for a helper method solution)*

That then causes trouble with the CronExpressionDescriptor used on the dashboard pages. It reads directly from the CRON string, which may not be the same as UTC. I'm in the Eastern U.S., so my CRON value is 4 hours ahead of the ExecutionTime.

jkodroff:
I am also on the East Coast of the US and getting a job to fire at 7:00 PM is trickier than expected.  Will post an update when we figure it out.

jkodroff:
Had to set the job to run at 6:59 PM (local) because the offset (-5 hours) would change the job to run on the wrong days.

jkodroff:
Had to set the job to fire at 6:59 PM because the offset (-5 hours) would then set the job to run on the wrong day.  Thus, I really don't like the UTC-only functionality.  It also makes it difficult for my users to read the admin panel (they don't really get UTC).

PR #327, opened 2015-03-17
Add time zone support for recurring jobs

odinserj:
This PR solves #160 and #270 by allowing to specify time zone for each recurring jobs. Each of the following jobs will fire at the same time (if Hawaii has no daylight saving time, I haven't google for this):

```csharp
RecurringJob.AddOrUpdate("UTC",      () => Console.WriteLine("UTC"),       "15 18 * * *");
RecurringJob.AddOrUpdate("UTC2",     () => Console.WriteLine("UTC"),       "15 18 * * *", TimeZoneInfo.Utc);
RecurringJob.AddOrUpdate("Russian",  () => Console.WriteLine("Russian"),   "15 21 * * *", TimeZoneInfo.Local);
RecurringJob.AddOrUpdate("Hawaiian", () => Console.WriteLine("Hawaiian"),  "15 08 * * *", TimeZoneInfo.FindSystemTimeZoneById("Hawaiian Standard Time"));
```

image

Here are some facts about current implementation, they will be moved to the docs:

* **Default time zone is UTC**. This may be confusing for users of Quartz.NET, Cron program in Linux, etc, that use *local time* by default. But switch to local time by default will introduce breaking changes, so it will be performed in version 2.0.
* **[Windows Time Zone IDs](https://msdn.microsoft.com/en-us/library/ms912391(v=winembedded.11).aspx)** are being used to store time zones. So there may be some issues when your Hangfire infrastructure contains both Windows and Linux. I thought about using [NodaTime](http://nodatime.org/) and its `DateTimeProviders.Tzdb` that uses [IANA time zone ids](http://en.wikipedia.org/wiki/List_of_tz_database_time_zones), but it is overkill. Custom time zone providers may be implemented for Hangfire to handle this issue later.
* **Issues with Mono**. `TimeZoneInfo` in Mono has some critical flaws in the latest stable Mono release at the time of this writing (2.10.8). `TimeZoneInfo.Local` has particular issues on the latest tested version of Mono. On Windows, `TimeZoneInfo.Local` throws `TimeZoneNotFoundException`. On Unix it returns a `TimeZoneInfo` with an Id of "Local", which isn't terribly useful (although it may contain the correct rules). – from http://nodatime.org/1.3.x/userguide/mono.html. This may be solved later, with custom time zone providers.

### DIFF
--- a/samples/ConsoleSample/ConsoleSample.csproj
+++ b/samples/ConsoleSample/ConsoleSample.csproj
@@ -38,9 +38,6 @@
     <StartupObject />
   </PropertyGroup>
   <ItemGroup>
-    <Reference Include="NodaTime">
-      <HintPath>..\..\packages\NodaTime.1.3.1\lib\net35-Client\NodaTime.dll</HintPath>
-    </Reference>
     <Reference Include="System" />
     <Reference Include="System.Core" />
     <Reference Include="System.Xml.Linq" />
@@ -57,7 +54,6 @@
   </ItemGroup>
   <ItemGroup>
     <None Include="app.config" />
-    <None Include="packages.config" />
   </ItemGroup>
   <ItemGroup>
     <ProjectReference Include="..\..\src\Hangfire.SqlServer.Msmq\Hangfire.SqlServer.Msmq.csproj">
--- a/samples/ConsoleSample/ConsoleSample.csproj
+++ b/samples/ConsoleSample/ConsoleSample.csproj
@@ -38,6 +38,9 @@
     <StartupObject />
   </PropertyGroup>
   <ItemGroup>
+    <Reference Include="NodaTime">
+      <HintPath>..\..\packages\NodaTime.1.3.1\lib\net35-Client\NodaTime.dll</HintPath>
+    </Reference>
     <Reference Include="System" />
     <Reference Include="System.Core" />
     <Reference Include="System.Xml.Linq" />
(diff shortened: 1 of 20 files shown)


odinserj:
Aaand unit test that is using `Hawaiian Standard Time` failed on Mono, because it does not know this time zone id (but it knows `Pacific/Honolulu`, `Pacific/Rarotonga`, etc.).

```
Hangfire.Core.Tests.Server.RecurringJobSchedulerFacts.Execute_GetsInstance_InAGivenTimeZone [FAIL]
   System.TimeZoneNotFoundException : Exception of type 'System.TimeZoneNotFoundException' was thrown.
   Stack Trace:
     at System.TimeZoneInfo.FindSystemTimeZoneByFileName (System.String id, System.String filepath) [0x00000] in <filename unknown>:0 
     at System.TimeZoneInfo.FindSystemTimeZoneById (System.String id) [0x00000] in <filename unknown>:0 
     at Hangfire.Core.Tests.Server.RecurringJobSchedulerFacts.Execute_GetsInstance_InAGivenTimeZone () [0x00000] in <filename unknown>:0 
     at (wrapper managed-to-native) System.Reflection.MonoMethod:InternalInvoke (System.Reflection.MonoMethod,object,object[],System.Exception&)
     at System.Reflection.MonoMethod.Invoke (System.Object obj, BindingFlags invokeAttr, System.Reflection.Binder binder, System.Object[] parameters, System.Globalization.CultureInfo culture) [0x00000] in <filename unknown>:0
```

PR #330, opened 2015-03-26
Outer context for job execution by UnitOfWorkManager.

itmagination:
`UnitOfWorkManager` provides outer context for job execution which is useful for scopes management (database transactions, DI container scopes etc.) and for externalization of any other kind of repetitive code on begin or end of each executed task.

To use `UnitOfWorkManager` you have to implement interface `IUnitOfWorkManager` and provide your implementation to Hangfire configuration. This is very similar to `JobActivator` usage.

For example to create `WindorContainer` scope for each job execution context you can use blow implementation:
```csharp
class WindsorContainerScopeUnitOfWorkManager : IUnitOfWorkManager
{
  private readonly IWindsorContainer _container;

  public WindsorContainerScopeUnitOfWorkManager(IWindsorContainer container)
  {
     _container = container;
  }

  public object Begin()
  {
    return _container.BeginScope();
  }

  public void End(object context, Exception ex = null)
  {
    ((IDisposable) context).Dispose();
  }
}
```
And register it in Hangfire like below:
```csharp
app.UseHangfire(config =>
{
    config.UseActivator(new WindsorJobActivator(_container.Kernel));
    config.UseUnitOfWorkManager(new WindsorContainerScopeUnitOfWorkManager(_container));
});
```

### DIFF
--- a/src/Hangfire.Core/BackgroundJobServer.cs
+++ b/src/Hangfire.Core/BackgroundJobServer.cs
@@ -21,6 +21,7 @@ using System.Linq;
 using Hangfire.Logging;
 using Hangfire.Server;
 using Hangfire.States;
+using Hangfire.UnitOfWork;
 
 namespace Hangfire
 {
@@ -146,6 +147,7 @@ namespace Hangfire
                 _storage,
                 new JobPerformanceProcess(),
                 JobActivator.Current,
+                UnitOfWorkManager.Current,
                 stateMachineFactory);
 
             yield return new WorkerManager(sharedWorkerContext, _options.WorkerCount);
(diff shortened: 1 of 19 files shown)


odinserj:
Hi @itmagination, sorry for the looong delay :frowning: In [Hangfire 1.5.0-beta1](https://github.com/HangfireIO/Hangfire/releases/tag/v1.5.0-beta1) I've added the `JobActivator.BeginScope` method as well as the base `JobActivatorScope` class to allow support for custom scopes/child containers without any job filters to simplify the user experience. Please see [my answer](http://discuss.hangfire.io/t/hangfire-simpleinjector-or-any-ioc/1147/4?u=odinserj) on a corresponding forum topic for some details.

I'm closing this PR. Thank you for it, you shown the way how to solve this problem using current features to many users! :heart:

PR #340, opened 2015-04-16
Fixed incompatibility issue with RabbitMQ.Client >=v3.4.0

justmara:
In RabbitMQ.Client v3.4.0 there was introduced some namespace changes making it incompatible with current code.
Simply updated package dependencies and code for it.

### DIFF
--- a/src/Hangfire.SqlServer.RabbitMq/Hangfire.SqlServer.RabbitMq.csproj
+++ b/src/Hangfire.SqlServer.RabbitMq/Hangfire.SqlServer.RabbitMq.csproj
@@ -38,9 +38,9 @@
     <NoWarn>1591</NoWarn>
   </PropertyGroup>
   <ItemGroup>
-    <Reference Include="RabbitMQ.Client, Version=3.3.0.0, Culture=neutral, PublicKeyToken=89e7d7c5feba84ce, processorArchitecture=MSIL">
+    <Reference Include="RabbitMQ.Client, Version=3.4.0.0, Culture=neutral, PublicKeyToken=89e7d7c5feba84ce, processorArchitecture=MSIL">
       <SpecificVersion>False</SpecificVersion>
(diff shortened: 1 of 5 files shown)


justmara:
doh. /me idiot. second commit wasn't planned for this pull request.

justmara:
this PR fixes the `Field not found: 'RabbitMQ.Client.ConnectionFactory.UserName'.` exception when connecting to RabbitMQ server 3.4+

odinserj:
Thanks, @justmara! This is a breaking change (updated dependency), however without it RabbitMQ Server 3.4 does not work with the old code, right? If yes, please re-create this PR to branch [release-1.4.2](https://github.com/HangfireIO/Hangfire/tree/release-1.4.2) if you want to see this release in 1.4.2, or to `dev` if it can wait for 1.5.0. 

If client update is not required, please roll back those changes please first.

odinserj:
@justmara, any news?

justmara:
Doh, sorry for not replying so long.
The main reason this PR was made is that I were having problems connecting to rabbit 3.4 with hangfire. I've tried updating rabbitmq.client package to latest version with no luck.
Then I took the sources and found that they simply wont compile with newer rmq lib. Found first incompatible version, made the fix, tried in my project and everything worked fine. And made pull request.
But then I've tried to rollback everything and.. it works fine with rabbitmq.client 3.5.1
So I simply don't have enough time to test all the cases and find what gone wrong at first try.

Looks like you can simply decline this PR or delay it for later ;)

odinserj:
Don't worry, your long replies can't be compared with my long replies :smile: Looks like you've made a comprehensive research, so this fix should be shipped ASAP. It's miserably that RabbitMQ client assembly is not ILMerged with Hangfire.SqlServer.RabbitMq assembly – in that case we could update it as an internal dependency without any breaking changes.

Given that, I'm suggesting to include this change to the upcoming 1.4.2 release, despite of breaking changes. 

@justmara, @dennyferra, what do you think about this?

Issue #350, opened 2015-04-28
Support for generic continuations


nirinchev:
It could very well be that I am missing something, but as far as I can see, currently, continuations don't allow the generic ```.ContinueWith<T>(string, Expression<Action<T>>)``` invocation. So it seems that the following is not supported:
```
var parent = BackgroundJob.Enqueue<MyBackgroundWorker>(worker => worker.DoTask1());
BackgroundJob.ContinueWith<MyBackgroundWorker>(parent, worker => worker.DoTask2());
```

If I misunderstood how continuations are to be used, don't hesitate to let me know :cat:

odinserj:
Omg, you are definitely right! Let's consider this as a "design bug" and add these methods to the 1.4.2 release. Would you like to create a PR for these methods for `release-1.4.2` branch?

nirinchev:
That depends on when you plan to release 1.4.2 :smile: I'm on a vacation until May 3rd, so if no one picks it up by then, I'll have a look.

odinserj:
Okay, have a good vacation!

yngndrw:
Created pull request: #356

I just noticed that the pull request is for master but you asked for it to go into release-1.4.2 - I'm new to GitHub but I'm guessing you can change it when you approve ? Let me know if you need me to do a new pull request.

yngndrw:
Created pull request: #357

odinserj:
Closing the issue – implemented in #357.

PR #357, opened 2015-05-02
Added support for generic continuations

yngndrw:
Added generic overloads for BackgroundJob.ContinueWith
Updated samples to use generic version of BackgroundJob.ContinueWith

See issue #350


odinserj:
Thanks, Andrew!

PR #364, opened 2015-05-10
Fetch recurring jobs using search criteria

akos-sereg:
With this change, users would be allowed to query Recurring Jobs by search criteria, where search criteria can be: RecurringJobID contains/starts with/ends with.

The purpose of the change: provided that we have millions of rows in HangFire.Set table, calling JobStorage.Current.GetConnection().GetRecurringJobs() and filtering results with Linq would not be performance efficient. Filtering at SQL level (this PR) is faster.

### DIFF
--- a/src/Hangfire.SqlServer/SqlServerConnection.cs
+++ b/src/Hangfire.SqlServer/SqlServerConnection.cs
@@ -253,14 +253,30 @@ where j.Id = @jobId";
             return new HashSet<string>(result);
         }
 
-        public override HashSet<string> GetAllItemsFromSet(string key, string valueFragment)
+        public override HashSet<string> GetAllItemsFromSet(string key, SearchCriteria criteria)
         {
 	        if (key == null) throw new ArgumentNullException("key");
+            if (criteria == null) throw new ArgumentNullException("criteria");
 
-	        string valueContains = string.Format("%{0}%", valueFragment);
+            string valuePattern = null;
+            switch (criteria.SearchMode)
+            {
+                case SearchMode.Contains:
+                    valuePattern = string.Format("%{0}%", criteria.Text);
+                    break;
+                case SearchMode.StartsWith:
+                    valuePattern = string.Format("{0}%", criteria.Text);
+                    break;
+                case SearchMode.EndsWith:
+                    valuePattern = string.Format("%{0}", criteria.Text);
+                    break;
+                default: 
+                    throw new ArgumentNullException(string.Format("Unsupported search mode: {0}.", criteria.SearchMode));
+            }
+
 	        var result = _connection.Query<string>(
-		        @"select Value from HangFire.[Set] where [Key] = @key and Value like @valueContains",
-		        new { key, valueContains });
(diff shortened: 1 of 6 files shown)


src/Hangfire.SqlServer/SqlServerConnection.cs
```diff
@@ -253,6 +253,34 @@ public override HashSet<string> GetAllItemsFromSet(string key)
...
+                    valuePattern = string.Format("%{0}%", criteria.Text);
+                    break;
+                case SearchMode.StartsWith:
+                    valuePattern = string.Format("{0}%", criteria.Text);
+                    break;
+                case SearchMode.EndsWith:
+                    valuePattern = string.Format("%{0}", criteria.Text);
+                    break;
+                default: 
+                    throw new ArgumentNullException(string.Format("Unsupported search mode: {0}.", criteria.SearchMode));
```
nirinchev: Not a big deal, but shouldn't this be ```InvalidEnumArgumentException ```?

akos-sereg: Sounds reasonable, changed to `InvalidEnumArgumentException`

odinserj:
Two years later, and I still can't find an abstraction with efficient implementation for each storage without any iteration through all the jobs. Definitely, SQL Server has better support for this and at least iterates in SQL Server itself. But when using Redis, to avoid awful iteration on client, we should have an index and maintain it manually. And since there are no notifications regarding key expiration, we also will need to add another index for expired jobs. And there will be another component for iterating expired jobs – a lot of complexity is added. 

So let's close this PR 😔 Maybe there's a way to get this work outside of Hangfire.

burningice2866:
Overkill solution would be to expose the `GetRecurringJobs()` as an `IQueryable`, then the storage provider would at least have a chance to defer the querying to its native capabilities while still maintaining the linq-experience in Hangfire.Core.

I did like the SearchCriteria class though as a simpler approach.

yngndrw-sage:
Redis does support server-side Lua scripting, so that could be an option to at least prevent iteration on the client.

Also since Redis 2.8, there are keyspace notifications which allow you to be notified of evictions and expiry: https://redis.io/topics/notifications (I've never used this feature, but it looks like it will do the job - You do still have to handle the case where a notification is missed, however)

odinserj:
@burningice2866, definitely overkill. I'm afraid of this interface and its implementation 😄 I see Hangfire as a minimum implementation to run all the background tasks. If someone wants to have better management opportunities, he or she always can build a layer on top of Hangfire that works best within their infrastructure. 

For example, we can create the table `RecurringJobs` in our application add fulltext index to it and insert records to that table before creating recurring jobs in Hangfire itself. Then add some interface for querying those jobs and trigger them. I agree this requires some more work rather then simply install Hangfire, but we can't make everyone happy.

odinserj:
@yngndrw, sooner or later we'll miss that notification, because pub/sub is never reliable when both publisher and subscriber are not in the same process. So sooner or later, we'll be required to rebuild an index from the ground up.

Issue #167, opened 2014-07-17
Should be able to specify seconds in RecurringJobScheduler


sgrassie:
The `RecurringJobScheduler` should be able to set a job to reoccur every X seconds.

There is support for this in NCrontab [see #7](https://code.google.com/p/ncrontab/issues/detail?id=7), although a build has not yet been released.

I think it would involve making a change [here](https://github.com/HangfireIO/Hangfire/blob/master/src/Hangfire.Core/RecurringJobManager.cs#L64), such that `CrontabSchedule.Parse` takes a second options argument, where you specify that you want to recognise six-part crontab expressions.

bbakkebo:
Hi, I am currently testing 1.40, but I cannot see that it was updated.  Hangfire works great, but I have one job that needs to run everything 30 seconds..Is there a workaround to this?
Thanks in advance. :+1:

bahram-hodjaev:
bbakkebo, you can workaround this

    public void Configuration(IAppBuilder app)
    {
        GlobalConfiguration.Configuration.UseSqlServerStorage("DefaultConnection");            
        app.UseHangfireDashboard();
        app.UseHangfireServer();
        BackgroundJob.Enqueue(() => Cron());
    ...

    public void Cron()
    {
        BackgroundJob.Schedule(() => Cron(), System.TimeSpan.FromSeconds(15));
    }

reggieboyYEAH:
Hi @bahram-hodjaev i tested this and this does not work,
this will only run the job once. 

any update on this? i have a job that needs to run every 30 seconds.

Thanks in advance :)

bahram-hodjaev:
Hi, @reggieboyYEAH, for now i am using this

    protected void Application_Start()
    {
        AreaRegistration.RegisterAllAreas();
        FilterConfig.RegisterGlobalFilters(GlobalFilters.Filters);
        RouteConfig.RegisterRoutes(RouteTable.Routes);
        BundleConfig.RegisterBundles(BundleTable.Bundles);

        clock = new ApplicationClock().Start();
    }   

    //...
    public class ApplicationClock
    {
        readonly object tickLock = new object();
        private Timer ticker = null;

        void tick(object _)
        {
            lock (tickLock)
            {
                try
                {
                    // DoSomeStuff();
                }
                catch
                {
                    if (ticker != null)
                    {
                        ticker.Dispose();
                    }

                    Start();
                }
            }
        }

        public ApplicationClock Start()
        {
[...]

PR #368, opened 2015-05-14
added support for seconds in NCronTab expressions

edokan:
added support for seconds in NCronTab expressions. 

Needed to update NCronTab and CronExpressionDescriptor to latest versions to ensure seconds part is compatible. Seconds support is kinda hacky.

Added some unit tests.


odinserj:
Does it work? The `RecurringJobScheduler` class is still waking up every minute.

edokan:
Actually, no. Well, partially. This patch allows, one to be able to specify seconds in crontab expressions, and my test are all about that. Actual work is to get scheduler to process them in seconds not in minutes as you explained. In my stupidest hour, I've assumed that should be enough, but it actually is not. Now I've enabled "seconds" in scheduler and some other places. It is less than ideal (SQL polling interval is now .5 seconds when second resolution is enabled), but It looks like it is working. Today I will create another pull request for that change. I'm new to all this pull request, so please forgive me.

edokan:
QuickDoc: GlobalConfiguration.Configuration.UseSecondResolution() to enable it

odinserj:
Don't worry, let's do it :smile:

odinserj:
Is there any way to lower the number of resulting classes without loosing their simplicity?

odinserj:
I think that second and minute throttling are the only cases, other invervals are unlikely be implemented.

odinserj:
Oh, @edokan, can you also re-target this feature against the `dev` branch? `master` is used only for bug fixes.

edokan:
Hi,

Here are my replies.

We had outsourced a project that uses hangfire for scheduled jobs, which is pretty nice, but they set so jobs are scheduled to work on every 15 minutes.

> What is the purpose of this change? I don't think it is necessary.

we needed some jobs to be scheduled to run on every 15 seconds, and there is an issue (#167) for that waiting for about a year.

> Is there any way to move SchedulerResolution dependency from ScheduleInstant to RecurringJobScheduler?

I think it is possible, but code may not be cleaner. Let me have a look.

> Is there any way to lower the number of resulting classes without loosing their simplicity?

It is possible, My first implementation used a lot less classes, then I refactored. Simplicity my not suffer that much

> I think that second and minute throttling are the only cases, other invervals are unlikely be implemented.

"minute" implementation covers 99.9% of cases, with "second" implementation, almost every case should be covered. I just did not want to leak to your implementation too much.

> Oh, @edokan, can you also re-target this feature against the dev branch? master is used only for bug fixes.

I will.

odinserj:
Ok, will wait for the changes. Regarding to the "What is the purpose of this change? I don't think it is necessary.", I've asked it because "every second" functionality should not cause to change the  `JobInitializationWaitTimeout`.

src/Hangfire.Core/SchedulerResolution.cs
```diff
@@ -0,0 +1,71 @@
...
+// License along with Hangfire. If not, see <http://www.gnu.org/licenses/>.
+
+using System;
+using Hangfire.Server;
+
+namespace Hangfire
+{
+    public abstract class SchedulerResolution
+    {
+        private static SchedulerResolution _current = new MinuteSchedulerResolution();
```
nirinchev: Not a big deal, but I don't like that the abstract class is aware of its inheritors.

odinserj:
Hi @edokan! Any news on this PR?

jmalczak:
Hi Guys. Is there any progress with that? I would like to use hangfire to pull amazon simple queue service every 20 seconds.

Moulde:
+1 :) We also have a need for less than minute recurring jobs.

lsoler:
+1 Hi! I would also like to use hangfire for recurring jobs every 30 seconds.

tukaef:
+1

odinserj:
Use background processes appeared in [Hangfire 1.5.0](https://www.hangfire.io/blog/2015/07/31/hangfire-1.5.0-beta.html#background-processes) to handle sub-minute intervals. Recurring jobs currently too slow to handle such intervals, since all of them are iterated on each run, harnessing the storage. Their implementation will be changed in Hangfire 2.0, opening a way to use seconds in intervals.

Jedarc:
Hello my friends, today I have an activity that is requiring recurring processes every 20 seconds. I know this item stayed for version 2.0, but is there any way I can implement it myself in my application?

burningice2866:
@Jedarc you can register a custom BackgroundProcess and set the delay inside it to 20 seconds. I do this for all my jobs which needs to run continuously/more often than ever minute.

Create a class that implements `IBackgroundProcess` and pass the instance of that class to `UseHangfireServer`

Jedarc:
@burningice2866 many thanks for the quick response, would you have any examples of this implementation, please?

burningice2866:
@Jedarc The interface has a single method where you implement your logic. 

The easiest is to wrap your code in a `while (!context.IsShutdownRequested){` block, and as the last thing in the block you call `context.CancellationToken.WaitHandle.WaitOne(TimeSpan.FromSeconds(20));`

So something like

```
public void Execute(BackgroundProcessContext context)
{
   var twentySeconds = TimeSpan.FromSeconds(20);

   while (!context.IsShutdownRequested)
   {
      ... your logic

      context.CancellationToken.WaitHandle.WaitOne(twentySeconds);
   }
}
```

PR #373, opened 2015-05-17
Allow to custom JobActivator for each server instance

fpellet:
I have a windows service with several hangfire servers. I like have several independent configurations.


odinserj:
Please see the https://github.com/HangfireIO/Hangfire/pull/374#issuecomment-124475247, I've posted it there by a mistake.

Issue #369, opened 2015-05-15
Sql Queue name length


maleet:
I like to have descriptive queue names and length of 20 is quite restrictive. Can't it be nvarchar(50)?

    [HangFire].[JobQueue](
	[Id] [int] IDENTITY(1,1) NOT NULL,
	[JobId] [int] NOT NULL,
	[Queue] [nvarchar](20) NOT NULL,
	[FetchedAt] [datetime] NULL)

I know I can have instead of queue named "company_client_web_result_jobs" "c_c_w_r_j" but then I need to have separate documentation what describes purpose of this queue.

odinserj:
Agreed, would you like to create a PR for this?

odinserj:
Closing the issue as PR #376 is merged into the `dev` branch.

PR #376, opened 2015-05-18
Increase SQL JobQueue.Queue column size

maleet:
Increased Queue name and Server Id lengths as discussed at #369

### DIFF
--- a/src/Hangfire.SqlServer/Install.sql
+++ b/src/Hangfire.SqlServer/Install.sql
@@ -133,7 +133,7 @@ BEGIN
         CREATE TABLE [HangFire].[JobQueue](
             [Id] [int] IDENTITY(1,1) NOT NULL,
             [JobId] [int] NOT NULL,
-            [Queue] [nvarchar](20) NOT NULL,
+            [Queue] [nvarchar](50) NOT NULL,
             [FetchedAt] [datetime] NULL,
             
             CONSTRAINT [PK_HangFire_JobQueue] PRIMARY KEY CLUSTERED ([Id] ASC)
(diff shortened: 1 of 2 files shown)


odinserj:
This will only increase column sizes for new installations. For old ones nothing will be changed, because the corresponding migration script will not be executed. You should revert your changes and add a new migration to the `install.sql` file with queries to modify existing tables instead to close the issue.

maleet:
With migration it's much better indeed.

src/Hangfire.SqlServer/Install.sql
```diff
@@ -344,13 +344,26 @@ BEGIN
 		SET @CURRENT_SCHEMA_VERSION = 4;
 	END
 
-	/*IF @CURRENT_SCHEMA_VERSION = 4
+	IF @CURRENT_SCHEMA_VERSION = 4
 	BEGIN
 		PRINT 'Installing schema version 5';
 
+		ALTER TABLE [HangFire].[JobQueue] ALTER COLUMN [Queue] NVARCHAR (50) NOT NULL;
+		PRINT 'Modified Queue length to 50';
```
odinserj: Please provide exact column names (with table name) for messages.

odinserj:
Okay, good, but some steps needs to be done in the crappy migration implementation:

* Update the `SET @TARGET_SCHEMA_VERSION = 4;` row in the top of the `install.sql` file.
* Update the `SqlServerObjectsInstaller` class in two places.

odinserj:
@maleet, thank you so much! Merging!

Issue #371, opened 2015-05-15
Hangfire tables are not case-sensitive


ilessa:
Hi, we're using a bin collation MSSQL database (necessary to support our customers) and recently received the following error (below) which we have isolated to the CountersAggregator class.

We require the table names to match the ones created (including case) so:

```sql
DELETE TOP (@count) [Hangfire].[Counter] with (readpast)
OUTPUT DELETED.[Key], DELETED.[Value], DELETED.[ExpireAt] INTO @RecordsToAggregate
SET NOCOUNT ON
MERGE [Hangfire].[AggregatedCounter] AS [Target]
```

would become

```sql
DELETE TOP (@count) [HangFire].[Counter] with (readpast)
OUTPUT DELETED.[Key], DELETED.[Value], DELETED.[ExpireAt] INTO @RecordsToAggregate
SET NOCOUNT ON
MERGE [HangFire].[AggregatedCounter] AS [Target]
```

Thanks.

```
Error occurred during execution of 'SQL Counter Table Aggregator' component. Execution will be retried (attempt 10 of 2147483647) in 00:01:28 seconds.
System.Data.SqlClient.SqlException (0x80131904): Invalid object name 'Hangfire.Counter'.
Transaction count after EXECUTE indicates a mismatching number of BEGIN and COMMIT statements. Previous count = 0, current count = 1.
   at System.Data.SqlClient.SqlConnection.OnError(SqlException exception, Boolean breakConnection, Action`1 wrapCloseInAction)
   at System.Data.SqlClient.SqlInternalConnection.OnError(SqlException exception, Boolean breakConnection, Action`1 wrapCloseInAction)
   at System.Data.SqlClient.TdsParser.ThrowExceptionAndWarning(TdsParserStateObject stateObj, Boolean callerHasConnectionLock, Boolean asyncClose)
   at System.Data.SqlClient.TdsParser.TryRun(RunBehavior runBehavior, SqlCommand cmdHandler, SqlDataReader dataStream, BulkCopySimpleResultSet bulkCopyHandler, TdsParserStateObject stateObj, Boolean& dataReady)
   at System.Data.SqlClient.SqlCommand.FinishExecuteReader(SqlDataReader ds, RunBehavior runBehavior, String resetOptionsString)
[...]

PR #378, opened 2015-05-20
Case sensitive changes for BIN collation SQL databases

ilessa:
Fixes for issue #371 - will work on all databases however changes are significant for users with a BIN collation SQL server.

### DIFF
--- a/samples/ConsoleSample/ConsoleSample.csproj
+++ b/samples/ConsoleSample/ConsoleSample.csproj
@@ -1,5 +1,6 @@
 ﻿<?xml version="1.0" encoding="utf-8"?>
 <Project ToolsVersion="4.0" DefaultTargets="Build" xmlns="http://schemas.microsoft.com/developer/msbuild/2003">
+  <Import Project="..\..\packages\xunit.runner.visualstudio.2.0.0\build\net20\xunit.runner.visualstudio.props" Condition="Exists('..\..\packages\xunit.runner.visualstudio.2.0.0\build\net20\xunit.runner.visualstudio.props')" />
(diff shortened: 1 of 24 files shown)


odinserj:
:heart: these changes. Why did you upgrade RabbitMQ client to 1.5.2?

ilessa:
I didn't have RabbitMQ on my machine and NuGet gave me 1.5.2.

odinserj:
Hm, strange, I've thought that NuGet installs the exact version specified in the project. Would you like to revert the `RabbitMq.Client` upgrade?

odinserj:
Btw, we can also create case sensitive database for unit-tests for more strict tests. Would you like to do this?

ilessa:
Ok, I think I've reverted the changes (first time I've been using GIT so bear with me)
Let me know if this is ok please.
I have the Hangfire.SqlServer.Tests database on my BIN collation server if that's what you mean?

odinserj:
Don't worry, all is ok :smile: `CleanDatabase` has a simple script to create database, and we can pass the CS collation:

```sql
CREATE DATABASE CaseSensitive COLLATE SQL_Latin1_General_CP1_CS_AS
```

ilessa:
Ok, I use Latin1_General_BIN if that’s any help

odinserj:
What do you mean regarding tempdb?

ilessa:
The tempdb database is built each time SQL Server is started, and has the same default collation as the model database, which is typically the same as the default collation of the instance. If you create a user database and specify a different default collation than model, the user database has a different default collation than tempdb. All temporary stored procedures or temporary tables are created and stored in tempdb, which means that all implicit columns in temporary tables and all coercible-default constants, variables, and parameters in temporary stored procedures have different collations than comparable objects created in permanent tables and stored procedures.

Just trying to prevent any collation conflicts if you use any temporary tables in the code :)

odinserj:
Time to merge this awesome PR! :heart:

Issue #379, opened 2015-05-20
Run recurrings jobs on per machine basis


bjornsallarp:
We have a load balanced application and have a couple of small jobs that fetches something externally and stores in memory on the server. Each job has to run on each machine, obviously, so I configured each server to have a unique queue and a shared default queue. The recurring jobs are attributed with  my "LocalQueueAttribute" to make them one per machine. The queue-name doesn't make the job unique from a job with the same name on a different queue so it doesn't work. Depending on which machine picks it up and schedules it will also run it.

Is there a way to achieve this already or functionality that could be added?

maleet:
+1

odinserj:
Since 1.5.0-beta1 it is possible to specify a queue for recurring job at runtime so you could use local queues for each machine. Please read http://hangfire.io/blog/#queues-for-recurring-jobs

bjornsallarp:
Awesome! I'll try it out. Thanks for the great work you're putting into this library!

PR #383, opened 2015-05-29
Run recurrings jobs on per machine basis #379

maleet:
For a bonus queue can be specified runtime.


odinserj:
Hi @maleet, thank you for this PR! I've merged it with some modifications: since recurring job manager does not process a job directly, the `queues` parameter in its ctor may be removed. Instead, we are creating a background job in enqueued state with the corresponding queue set:

```csharp
// ... from RecurringJobScheduler.cs
var state = new EnqueuedState { Reason = "Triggered by recurring job scheduler" };
if (recurringJob.ContainsKey("Queue") && !String.IsNullOrEmpty(recurringJob["Queue"]))
{
    state.Queue = recurringJob["Queue"];
}

var jobId = CreateBackgroundJob(connection, job, state, stateMachineFactory);
```

PR #384, opened 2015-05-29
Fixing bug in history when duration > int.MaxValue

mavenius:
Fixing bug in history when duration is longer than int.MaxValue ms.


odinserj:
Mavenius, thanks for the PR! I've merged the changes into the `dev` branch (697f1931c2c427d8200e1362799329a4b86fab29).

PR #396, opened 2015-07-07
Hangfire.SQLite added to related projects

vip32:


odinserj:
Wow, good work, @vip32. Merging!

odinserj:
Would you like to update the [extensions page](https://github.com/HangfireIO/HangfireIO.github.io/blob/master/extensions.md) as well?

PR #412, opened 2015-08-05
Fixes possible crash on Queue Page

erictrade:
At some circumstances, hangfire can create a "phantom job", that just have an jobId and doesn't have any job information. I couldn't track where it's creating it yet (I'm still using the old Unsupported Redis, so it may be already fixed in Pro or SQL Server versions).
The problem is that this "phantom job" may crash the queue page, with a NullReference:

image

This pull request purpose a protection to these specific cases. What do you think?


odinserj:
Hi @erictrade, I've just merged your commit to the `dev` branch – 3fb31cbedcdc07d2fc2fa925128c55498cdb6acb. Thank you!

erictrade:
Wow, that was fast! I'm happy to contribute!

erictrade:
Hi, I would like to inform that probably these "phantom jobs" are related to the recurring jobs. Once all recurring jobs enqueued were consumed, the page stopped crashing, even before using this NullReference protection. But I still doesn't know what exactly is creating these jobs without information.
Maybe, when a recurring job is enqueued, some auxiliary job is created to control something on the process? Didn't analysed id deeply, just wondering.
If I can dig some more information, I'll report here, and if I have enough data, I'll open an issue, or maybe try to fix.

PR #413, opened 2015-08-10
Add DisplayName property to Job

barclayadam:
Allows overriding the name of a job directly, in addition to the currently supported DisplayNameAttribute on the executed method.

With a custom creation of a job that uses a single generic method all of the jobs have the same name. This allows us to override that for each job instance that is created, instead of each job 'type'

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -2,7 +2,7 @@
 # see travis-ci.org for details
 
 language: csharp
-sudo: false
+sudo: required
 
 # Make sure build dependencies are installed.
 install:
--- a/Hangfire.sln
+++ b/Hangfire.sln
@@ -1,7 +1,7 @@
 ﻿
 Microsoft Visual Studio Solution File, Format Version 12.00
 # Visual Studio 2013
-VisualStudioVersion = 14.0.22823.1
+VisualStudioVersion = 12.0.31101.0
 MinimumVisualStudioVersion = 10.0.40219.1
 Project("{FAE04EC0-301F-11D3-BF4B-00C04F79EFBC}") = "ConsoleSample", "samples\ConsoleSample\ConsoleSample.csproj", "{C02BB718-2AE4-434C-8668-C894FF663FCE}"
 EndProject
--- a/Hangfire.sln
+++ b/Hangfire.sln
@@ -1,7 +1,7 @@
 ﻿
 Microsoft Visual Studio Solution File, Format Version 12.00
 # Visual Studio 2013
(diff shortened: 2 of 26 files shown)


odinserj:
Hi @barclayadam! When do you use them, why `DisplayNameAttribute` does not satisfy your needs? Can you paste here a screenshot?

barclayadam:
Hi @odinserj. This follows on from the on-going discussion about needing `TypeNameHandling.All`. I have a single, central way of executing background tasks: 

    public class TaskExecutor {
        public void Execute(ITask task) { ... }
    }

Because of this *every* task when executed has a name of `TaskExecutor.Execute`, which tells me nothing of the actual task. I cannot have a `DisplayNameAttribute` on the method as it is not static, it depends on the type of task being executed. I can only determine at run time, not build time, what the name of the job should actually be.

When en-queuing a task I then use the manual method of creating a job (or creating via. `RecurringJobManager`:

    new BackgroundJobClient(JobStorage.Current).Create(TaskAsJob.Create(task), new EnqueuedState());

where `TaskAsJob.Create` looks like:

    public static Job Create(ITask task)
    {
        var job = Job.FromExpression<TaskExecutor>(e => e.Execute(task));
        job.DisplayName = "Execute(" + task.GetType().Name + ")";

        return job;
    }

This means I get a dashboard that looks like:

hangfire displayname

barclayadam:
Hey @odinserj 

I have rebased this pull request to the latest in master. It would be really handy for us to have this included as it is used extensively in our application.

The changes are pretty minimal and completely optional from an end-user point of view. I believe it complements the existing ability to provide a display name for a job for advanced scenarios

nirinchev:
Perhaps you could revert the .sln file? Or at least the portions that didn't change. Other than that, it does seem minor change, although I don't know if it aligns with @odinserj's vision of the product. To me it seems fairly niche use case. Furthermore, if you're using it for Recurring jobs only, you could set the id to a meaningful name, can't you?

barclayadam:
@nirinchev good call on the solution file, that had slipped in to this PR accidentally, have sorted it out now.

I do not use this just for recurring jobs, but also for 'normal' scheduled tasks. I am coming from a legacy system, which means we already had a concept of background tasks. Instead of rewriting everything I wanted to leverage our existing infrastructure as much as possible.

I agree this may be a niche case, but it enables configuration that may be required as the lower level bits of Hangfire are used, instead of the higher level abstractions (which I would imagine most people would use)

Issue #418, opened 2015-08-16
Stack trace markup code missing copyright notice


atifaziz:
The [stack trace markup code in Hangfire](https://github.com/HangfireIO/Hangfire/blob/5a770aab402b820f262f0872b18c2d3304bb7128/src/Hangfire.Core/Dashboard/HtmlHelper.cs#L250-L395) seems to [come from the ELMAH project](https://bitbucket.org/project-elmah/1x/src/7576b913e220d9ddf5473f5875f5ea8ce3380c9e/src/Elmah/ErrorDetailPage.cs?at=default#ErrorDetailPage.cs-383). ELMAH uses [Apache License, Version 2.0](https://bitbucket.org/project-elmah/1x/src/7576b913e220d9ddf5473f5875f5ea8ce3380c9e/COPYING.txt?at=default), which requires (at the very least and among other things) that the [copyright notice](https://bitbucket.org/project-elmah/1x/src/7576b913e220d9ddf5473f5875f5ea8ce3380c9e/src/Elmah/ErrorDetailPage.cs?at=default#ErrorDetailPage.cs-4) be maintained.

odinserj:
Hello, @atifaziz. I've added all the notices with commit 5ed91f13b37aba991b2977615b090afbf8716bef. I've also included them into [Third party files and licenses](http://hangfire.io/licensing.html#third-party-files-and-licenses) section in the official site. I'm adding all the used projects to give appropriate attributions, even when a license don't require to do so, but completely missed your code, sorry for this. Your stack trace highlighter significantly improves the debugging experience when something goes wrong, thank you for it!

Should I do anything else?

atifaziz:
@odinserj Thanks for addressing this so quickly and [dedicating a page where you explicitly list the various libraries and licenses](http://hangfire.io/licensing.html#third-party-files-and-licenses) is commendable and goes a step further than most.

> Should I do anything else?

I would've said that it's important to also mention that the stack trace highlighting code is licensed under [Apache License, Version 2.0](http://www.apache.org/dev/apply-license.html). Your modifications, if you had made any, would go with the terms of [LGPL v3](http://www.gnu.org/licenses/lgpl-3.0.en.html). However, I have a better proposal. Incidentally, I recently refactored and extracted the stack trace parsing and formatting code from [ELMAH](https://elmah.github.io/) into stand-alone projects and NuGet packages:

* [StackTraceParser](https://github.com/atifaziz/StackTraceParser)
[...]

PR #419, opened 2015-08-17
Use StackTraceFormatter for marking-up stack traces

atifaziz:
Instead of including stack trace highlighting code from ELMAH, use the more up-to-date and stand-alone version via the StackTraceFormatter package. Addresses issue #418 and also makes sync'ing easier going forward.


atifaziz:
Note that [StackTraceFormatter](https://github.com/atifaziz/StackTraceFormatter) code was slightly modified after importing [the package](https://www.nuget.org/packages/StackTraceFormatter.Source/1.0.0). Specifically, [lines 46-56](https://github.com/atifaziz/Hangfire/blob/6009543b204a8b8d0c3ad59cab3b8d1b255b90c8/src/Hangfire.Core/App_Packages/StackTraceFormatter/StackTraceFormatter.cs#L46-L56) in the original code were written as C# 6 expression-bodied function members and I has to change these to be compatible with C# 5.

odinserj:
Thank you for your time, @atifaziz!

PR #420, opened 2015-08-17
Stack trace parameters CSS class typo

atifaziz:
The typo had no impact so far as `st-params` was not styled in [`hangfire.css`](https://github.com/HangfireIO/Hangfire/blob/dev/src/Hangfire.Core/Dashboard/Content/css/hangfire.css#L186).


odinserj:
Thanks, Atif!

Issue #421, opened 2015-08-17
Worker does not respect JobFilters during StateMachine.ChangeState call


yngndrw:
When the worker picks up a job and changes its state from ```EnququedState``` to ```ProcessingState``` (https://github.com/HangfireIO/Hangfire/blob/5a770aab402b820f262f0872b18c2d3304bb7128/src/Hangfire.Core/Server/Worker.cs#L70) it correctly handles the case where a job cannot be loaded but does not correctly handle the case where a filter has changed the candidate state during the state election process. (https://github.com/HangfireIO/Hangfire/blob/5a770aab402b820f262f0872b18c2d3304bb7128/src/Hangfire.Core/States/StateMachine.cs#L127)

What actually happens in this case is that the job's state is changed to the new candidate but the worker still continues to "perform" the job. After the worker has "performed" the job, it attempts to change the job's state to ```SucceededState``` but as the current job state doesn't match the from state condition (```ProcessingState```) it does not get marked as completed as you might expect.

I'm thinking there's three changes that should be made in regards to this:
1) Within ```StateMachine.ChangeState(StateContext context, IState toState, string oldStateName)```, return the new state that has actually been set.
2) Within ```StateMachine.ChangeState(string jobId, IState toState, string[] fromStates, CancellationToken cancellationToken)```, return false if either the job data has not loaded successfully or if the inner ```ChangeState``` call has changed the state to a different state than was requested.
3) Within ```Worker.Execute(BackgroundProcessContext context)```, remove the from state conditions in the final ```ChangeState``` call as the job should always be changed to either ```SucceededState``` or ```FailedState``` if it has been executed. (Although a job filter should also be able to retrospectively change this afterwards.)

odinserj:
Yes, I've spotted the same problem a while ago and realized to return an elected state instead of a bool value: its `false` value says that the *failed* state was elected, and its `true` value should mean (but this isn't true now) that elected state equals to the initial one. And we can return `null`, if a state change process hasn't met preconditions.

Currently, the return value is used only by the `Worker` class, when moving a job to the *processing* state. And in my opinion, it should simply remove job from a queue without processing, if one of job filters elected another state. What do you think?

odinserj:
Btw, what state elects your job filter instead of processing? What do you want to perform?

yngndrw:
Just to clarify on the first part of your response, are you suggesting that the public ```StateMachine.ChangeState``` method should return a nullable boolean:
[...]

PR #422, opened 2015-08-18
The worker now correctly handles the case where it tries to change a …

yngndrw:
The worker now correctly handles the case where it tries to change a job to the processing state but a job filter changes the state to a different state

See issue #421

### DIFF
--- a/src/Hangfire.Core/IBackgroundJobClient.cs
+++ b/src/Hangfire.Core/IBackgroundJobClient.cs
@@ -51,7 +51,7 @@ namespace Hangfire
         /// <param name="jobId">Identifier of job, whose state is being changed.</param>
         /// <param name="state">New state for a job.</param>
         /// <param name="fromState">Current state assertion, or null if unneeded.</param>
-        /// <returns>True, if state change succeeded, otherwise false.</returns>
+        /// <returns>True, if state change succeeded and a job filter did not request another state, otherwise false.</returns>
         bool ChangeState(string jobId, IState state, string fromState);
     }
 }
--- a/src/Hangfire.Core/BackgroundJobClient.cs
+++ b/src/Hangfire.Core/BackgroundJobClient.cs
@@ -110,7 +110,9 @@ namespace Hangfire
             {
                 var factory = _stateMachineFactoryFactory.CreateFactory(_storage);
                 var stateMachine = factory.Create(connection);
-                return stateMachine.ChangeState(jobId, state, fromState != null ? new[] { fromState } : null) != null;
+                IState appliedState = stateMachine.ChangeState(jobId, state, fromState != null ? new[] { fromState } : null);
+
+                return appliedState != null && appliedState.Name.Equals(state.Name, StringComparison.OrdinalIgnoreCase);
             }
         }
     }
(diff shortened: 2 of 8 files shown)


yngndrw:
Travis CS build failed due to a timeout in some of the RabbitMq tests:
```
Test assembly: /home/travis/build/HangfireIO/Hangfire/tests/Hangfire.SqlServer.RabbitMq.Tests/bin/Release/Hangfire.SqlServer.RabbitMq.Tests.dll
Hangfire.SqlServer.RabbitMq.Tests.RabbitMqJobQueueFacts.DisposeWithoutRemoval_OnFetchedJob_CanBeDequeuedAgain [FAIL]
   Test execution time exceeded: 30000ms
Hangfire.SqlServer.RabbitMq.Tests.RabbitMqJobQueueFacts.RemoveFromQueue_OnFetchedJob_RemovesTheJobCompletely [FAIL]
   Test execution time exceeded: 30000ms
Hangfire.SqlServer.RabbitMq.Tests.RabbitMqJobQueueFacts.Dequeue_ReturnsFetchedJob_FromOtherQueues_IfFirstAreEmpty [FAIL]
   Test execution time exceeded: 30000ms
30 total, 3 failed, 0 skipped, took 95.651 seconds
```

src/Hangfire.Core/IBackgroundJobClient.cs
```diff
@@ -51,7 +51,7 @@ public interface IBackgroundJobClient
         /// <param name="jobId">Identifier of job, whose state is being changed.</param>
         /// <param name="state">New state for a job.</param>
         /// <param name="fromState">Current state assertion, or null if unneeded.</param>
-        /// <returns>True, if state change succeeded, otherwise false.</returns>
+        /// <returns>True, if state change succeeded and a job filter did not request another state, otherwise false.</returns>
```
odinserj: > True, if state change succeeded and a job filter did not request another state

This isn't true for the following code.

```csharp
return stateMachine.ChangeState(jobId, state, fromState != null ? new[] { fromState } : null) != null;
```

Should we replace it with the following line?

```csharp
return stateMachine.ChangeState(jobId, state, fromState != null ? new[] { fromState } : null) != fromState;
```

odinserj:
Other changes are very nice! Will be happy to merge this :heart:

yngndrw:
Good catch, I'll update it when I get back. I implemented it to try and be backwards compatible with the previous public API but I think it's best to expose the new functionality.

I'll update it with a check on the state's name as I have done in the worker rather than the specific instance of the state as a job filter may return an equivalent replacement state.

yngndrw:
Corrected the error that you spotted, thanks.

odinserj:
Thanks, Andrew! I'm merging this PR. Since it adds a breaking change anyway (will require to change Hangfire.Pro), I'll refactor the state change subsystem to make it looks like client and server.

odinserj:
@yngndrw, I've slightly changed the return value in PR #431. It now returns `FailedState` instance instead of `null` value on job load exception. `null` value is returned only when the corresponding job wasn't found.

yngndrw:
Thank you for the heads-up on this change Sergey, I think that makes more sense than how I left it. I've left a note on the XML comment for this method as it's now incorrect.

I haven't had a chance to do the quota changes yet but I think that's a good thing given the other state machine changes that you have just made. Now that those changes are merged in I can continue on with the quota functionality when I get the time.

odinserj:
Not a problem, Andrew. I've also released [1.5.0-beta3](http://hangfire.io/blog/2015/08/31/hangfire-1.5.0-beta3.html) with that change.

PR #423, opened 2015-08-19
Hide delete buttons

Cussa:
I made a implementation to hide the delete buttons from Dashboard. It's not based on user or role, but show or hide all the delete buttons. This because I implemented MEF integration to execute plugins from zip files, and want to delete only from the grid where you can remove both the zip files and the hangfire job.


odinserj:
Closing this as #934 covering this feature is merged

PR #428, opened 2015-08-27
Disable concurrent execution command timeout

Tom-Kelly:
Resolved an issue where the timeout for the Stored Procedure sp_getapplock could be longer than the default timeout on the Command object. This could result in the command timing out, not the SP call, and so an unexpected exception being thrown. 

The command object now has its timeout set to be 1 second longer than the SP timeout.

### DIFF
--- a/src/Hangfire.SqlServer/SqlServerDistributedLock.cs
+++ b/src/Hangfire.SqlServer/SqlServerDistributedLock.cs
@@ -57,7 +57,7 @@ namespace Hangfire.SqlServer
             parameters.Add("@LockTimeout", timeout.TotalMilliseconds);
             parameters.Add("@Result", dbType: DbType.Int32, direction: ParameterDirection.ReturnValue);
 
-            // Ensuring the timeout for the connection is 1 second longer than the timeout specified for the stored procedure.
+            // Ensuring the timeout for the command is 1 second longer than the timeout specified for the stored procedure.
             var connectionTimeout = timeout.Seconds + 1;
 
             connection.Execute(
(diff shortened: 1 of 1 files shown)


odinserj:
Thank you, @Tom-Kelly, I thought about this, but forgot to solve this problem. Congrats with your first pull request on GitHub!

odinserj:
I'll release this changes with [1.4.6](https://github.com/HangfireIO/Hangfire/milestones/1.4.6) coming tomorrow.

src/Hangfire.SqlServer/SqlServerDistributedLock.cs
```diff
@@ -44,6 +44,7 @@ public SqlServerDistributedLock(string resource, TimeSpan timeout, IDbConnection
         {
             if (String.IsNullOrEmpty(resource)) throw new ArgumentNullException("resource");
             if (connection == null) throw new ArgumentNullException("connection");
+            if (timeout.TotalSeconds > Int32.MaxValue) throw new ArgumentException("The timeout specified is greater than Int32.MaxValue when expressed as seconds.", "timeout");
```
yngndrw: Should this check be ```timeout.TotalSeconds >= Int32.MaxValue```, as 1 is later added before it is case to an Int 32 ?

odinserj: Correct, Andrew. I've already released this change with [1.4.6](http://hangfire.io/blog/2015/08/29/hangfire-1.4.6.html), but this isn't bad. @Tom-Kelly would you like to fix this by making a PR against the `dev` branch?

Tom-Kelly:
@odinserj @yngndrw Whoops. Yes thank you, I'm currently away without a laptop but I'll submit a PR to the dev branch on 3/9.

src/Hangfire.SqlServer/SqlServerDistributedLock.cs
Tom-Kelly: @odinserj @yngndrw Whoops, yes thank you. I'm currently away without a laptop but I'll submit a PR for this on Thursday, 3/9/15.

odinserj: @Tom-Kelly, okay. Have a nice laptop-free vacation :smile:

PR #431, opened 2015-08-29
State machine refactoring

odinserj:
`Hangfire.States` namespace was refactored to fix different architecture mistakes. `Client` and `Server` namespaces were also touched a bit, since they depend on a state machine mechanics. The following list contains the main differences, please read commit messages for the details.

* `IStateChangeProcess` is now used for changing a background job state using its id, when other details aren't available (it was the responsibility of the `IStateMachine` interface earlier).
* `IStateMachine` interface is now responsible for electing and applying a state to a background job. There is no more separate methods for electing and applying a state. State election process is carrying out inside a transaction.
* `ApplyStateContext` and `ElectStateContext` classes have `Connection` and `Transaction` properties available in state filters.
* `IJobCreationProcess` interface was simplified by removing the `IJobCreator` parameter (interface was removed also).
* Removed `IStateMachineFactory` and `IStateMachineFactoryFactory` interfaces :tada:.
* Added `Storage` property to `CreateContext` and `ApplyStateContext`, their constructors were changed.

These changes greatly improved the `Hangfire.States` subsystem. As always with better programming models, this resulted to many enhancements, including simplified background processes (no more factories) and better reliability for state filters (apply state filters can now query the data, this allow to prevent some theoretical race conditions).

### Breaking Changes

Almost each change is a breaking one, however these changes touch only low-level APIs. I've intentionally rolled back any changes that break extensions based on a documented high-level APIs (storages, filters, etc.), so existing extensions will be unlikely touched. 

However, new version of the `Hangfire.Pro` package will be released as it uses some internal mechanics, but the new state machine will make it more reliable.


src/Hangfire.Core/States/StateChangeProcess.cs
```diff
@@ -0,0 +1,150 @@
...
+        {
+            if (stateMachine == null) throw new ArgumentNullException("stateMachine");
+            _stateMachine = stateMachine;
+        }
+        
+        /// <summary>
+        /// Attempts to change the state of a job, respecting any applicable job filters and state handlers
+        /// <remarks>Also ensures that the job data can be loaded for this job</remarks>
+        /// </summary>
+        /// <returns><c>Null</c> if a constraint has failed or if the job data could not be loaded, otherwise the final applied state</returns>
```
yngndrw: This comment does not represent the functionality as ```FailedState``` will be returned on ```JobLoadException``` rather than ```null```. (Unless ```IgnoreJobLoadException``` is set.)

odinserj: Thanks, Andrew! Fixed in the 23fd4d18a46ed4e772d2f496f43f3a0beef9950c commit.

PR #434, opened 2015-09-03
SqlServerDistributedLockTimeout - logical error in value check

Tom-Kelly:
Previously the check was made for only `value > Int32.Max` but then 1 was added later. Now checking to see if `value >= Int32.Max`. 

Test also added for this.


odinserj:
I think it is better to introduce a constant and use it instead of a hard-coded value `1` in the following line and also use this const in a validation logic.

```csharp
var commandTimeout = (int)(timeout.TotalSeconds + 1);
```

The `>=` operator is only valid for `1` value, making the validation logic brittle.

Tom-Kelly:
@odinserj Good point, I'll do that.

odinserj:
Thanks, Tom! Merging the changes

PR #448, opened 2015-09-25
#1 Commit

MarceloBematech:
Install.sql converted to Oracle and Entities folder added


odinserj:
Nice work, @MarceloBematech! Please, use separate repository for Hangfire.Oracle, as other projects like Hangfire.PostgreSql or Hangfire.MySql to keep main repository as clean as possible.

PR #451, opened 2015-09-30
Simplify writing of embedded resource to response body.

robvdlv:
'Stream.CopyTo' introduced in .NET 4.5


odinserj:
Hm, it's really simpler. Thank you, Rob!

PR #466, opened 2015-10-23
Allow jobs to log directly to the hangfire dashboard

samshiles:
I have augmented parameter substitution in order to allow access to the job id in background jobs. This allows us to do the following:

  BackgroundJob.Enqueue<IRenewalJob>(x => x.Execute(JobContext.Null));

Hangfire will now substitute this with an instance of JobContext, allowing access to the jobId for logging purposes. A simple logger that makes use of this can be found at: 
https://github.com/samshiles/hangfire.job.logging

### DIFF
--- a/src/Hangfire.Core/JobContext.cs
+++ b/src/Hangfire.Core/JobContext.cs
@@ -1,0 +1,25 @@
+﻿using Hangfire.Storage;
+
+namespace Hangfire
+{
+    public class JobContext : IJobContext
+    {
+        private readonly string jobId;
+        private readonly string serverId;
+
+        public JobContext(string jobId, IStorageConnection connection)
+        {
+            this.jobId = jobId;            
+        }       
+
+        public string JobId
+        {
+            get { return jobId; }
+        }
+
+        public static IJobContext Null
+        {
+            get { return null; }
+        }
+    }
+}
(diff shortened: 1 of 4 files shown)


burningice2866:
+1

everfrown:
+1

burningice2866:
If not for the logging, then just to be able to retrieve information about the job, from with the job itself.

I know there are several hacks you can do regarding Thread Static properties, but this seems much cleaner and baked in.

rdehuyss:
+1 - great for logging in distributed systems

src/Hangfire.Core/JobContext.cs
```diff
@@ -0,0 +1,25 @@
+﻿using Hangfire.Storage;
+
+namespace Hangfire
+{
+    public class JobContext : IJobContext
+    {
+        private readonly string jobId;
+        private readonly string serverId;
```
celluj34: `serverId` is never used.

src/Hangfire.Core/JobContext.cs
```diff
@@ -0,0 +1,25 @@
...
+
+        public JobContext(string jobId, IStorageConnection connection)
```
celluj34: `connection` is never stored

mirhagk:
Is there plans to merge this into a release soon? We'd like to get this sometime in the next few months and we're willing to wait for an official release if it's coming soon. Otherwise we'll build from source

alastairtree:
+1

samiraguiar:
@odinserj Could you please comment on what could be improved/changed to make this PR be accepted? We really needed this.

thebodaciousd:
+1 Would love to see this merged

odinserj:
Guys, your dreams come true, please see [Hangfire.Console](https://github.com/pieceofsummer/Hangfire.Console) extension by @pieceofsummer. It adds support for logging, live updates and even progress bars! All you need is to install the extension and call the `UseConsole` method, please see the project's readme file.

image

PR #469, opened 2015-10-31
Prevent infinite loops when trying to change state of a job that doesn't exist

### LINKED ISSUE #467: Infinite loop when trying to delete job that doesn't exist
Hangfire version 1.5.2
I'm trying to delete job, that does not exist and my program goes into infinite loop. 
Is it possible, that the latest changes in BackgroundJobStateChanger.GetJobData method have caused this problem?

mkravchuk7:
Simple solution to break infinite loop and resolve #467.

This infinite loop happens when CancellationToken.None is used in one of the constructor overloads of 'StateChangeContext' class. 

Also maybe it is better to return null in 'GetJobData' method immediately after first try without any retries.

### DIFF
--- a/src/Hangfire.Core/States/BackgroundJobStateChanger.cs
+++ b/src/Hangfire.Core/States/BackgroundJobStateChanger.cs
@@ -132,7 +132,8 @@ namespace Hangfire.States
                     return jobData;
                 }
 
-                if (context.CancellationToken.IsCancellationRequested)
+                if (context.CancellationToken.IsCancellationRequested ||
+                    context.CancellationToken == CancellationToken.None)
                 {
                     return null;
                 }


mkravchuk7:
@odinserj 
Hi. Is #467 already fixed? Should I implement a different solution to fix it?

odinserj:
Hi @mkravchuk7 and sorry for the delay. Started to investigate this, the change is simple, but I should review it carefully, because it may cause different issues. I'll let you know the results!

odinserj:
Checked worker, delayed scheduler, background job client and continuations – looks fine. Thank you for the PR!

mkravchuk7:
Thank you for Hangfire:)

PR #477, opened 2015-11-10
Fix 404 for quickstart reference

dizzzan:


odinserj:
Thanks for the fix, @dizzzan! There was a redirect, but it isn't working anymore :-(

PR #479, opened 2015-11-11
SQL Server 2005

harriyott:
Tentative changes to allow Hangfire to work with SQL Server 2005.
- Changed `datetime2` column to `datetime`
- Changed SQL `merge` commands to separate `insert` and `update` commands
- Removed the `forceseek` table hints

Whilst the SQL unit tests pass, and my installation works with a simple background job, I am wary of any side effects that this PR may cause.

### DIFF
--- a/src/Hangfire.SqlServer/CountersAggregator.cs
+++ b/src/Hangfire.SqlServer/CountersAggregator.cs
@@ -31,13 +31,16 @@ namespace Hangfire.SqlServer
 
         private readonly SqlServerStorage _storage;
         private readonly TimeSpan _interval;
+        private readonly ISqlServerSettings _sqlServerSettings;
 
-        public CountersAggregator(SqlServerStorage storage, TimeSpan interval)
+        public CountersAggregator(SqlServerStorage storage, TimeSpan interval, 
+            ISqlServerSettings sqlServerSettings = null)
         {
             if (storage == null) throw new ArgumentNullException("storage");
 
             _storage = storage;
             _interval = interval;
+            _sqlServerSettings = sqlServerSettings;
         }
 
         public void Execute(CancellationToken cancellationToken)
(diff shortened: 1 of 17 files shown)


odinserj:
Hi @harriyott, nice PR! How do you think, what side effects this PR may cause?

harriyott:
Glad you like it. I had two concerns; firstly, the `holdlock` in the merge - I haven't done anything equivalent in the replacement code. Secondly, I was concerned that removing `forceseek` may have side effects. 

I guess if there could be problems, then a check could be done for v2005 which runs my code, and newer versions run the original.

odinserj:
Okay, we have 3 groups of changes:

* **`merge`** statement replaced with `update`/`insert`. The latter operations in `SqlServerConnection` class should be atomic and so wrappen in a transaction. This replacement will also only work correctly with `serializable` isolation level, because a row with the same key may be added between the two commands. Hangfire allows to change the isolation level, so this may cause incorrect behavior.
* **`forceseek`** hint is removed. With large amount of data, with millions of background jobs, this will cause a significant slowdown. This hint is here for a reason.
* **`datetime2`** is replaced with `datetime`. Hm, this is OK, however there are a lot of people who already has `datetime2` in their real tables. Are we really can just change it in a migration script?

harriyott:
Yes, you're right. All of these changes would be harmful to those using newer versions, so the version should be checked. It probably needs a bit of restructuring if you want to include proper support for 2005.

odinserj:
How do you plan to support it? There may be `SqlServerStorageOptions.CompatibilityMode` setting or something like this.

odinserj:
Oh, looks like it's time to close this PR – SQL Server 2008, 2012, 2014 and 2016 enough to support for me 😄  And there's a way to create an additional package for SQL Server 2015.

PR #481, opened 2015-11-11
Separated SQL Server 2005 code from newer versions

harriyott:
- Created a new setting for SQL Server 2005
- New interface and two classes for differences in SQL for 2005 / newer
- Check for setting and get correct SQL in places where version distinctions are required


odinserj:
There is no need for a new PR, an old one (https://github.com/HangfireIO/Hangfire/pull/479) is automatically updated when you push new changes to a branch. So I'm closing this, an will comment an old one asap.

PR #485, opened 2015-11-20
Make SqlServerObjectsInstaller public

vbfox:
We have the need to run the update the HangFire during our database update / release process. The only current way is to create an instance of `SqlServerStorage` that does a bunch of other things or call the method via reflection.

This PR represent the simplest change to make it happen but i'm open to any other method as long as we can initialize the database separately from the Hangfire startup.


odinserj:
I've merged this PR into the `dev` branch. Please see the commit bd77a8d74ce92d411c1b66576a66e29a10458dfc. Thank you for the PR! The feature will be released in 1.6.0.

PR #491, opened 2015-12-16
IRecurringJobManager interface for RecurringJobManager added

arnoldasgudas:
It would be really great to have interface for RecurringJobManager so that it would be able to mock it in  unit tests.

### DIFF
--- a/src/Hangfire.Core/IRecurringJobManager.cs
+++ b/src/Hangfire.Core/IRecurringJobManager.cs
@@ -1,0 +1,18 @@
+﻿using System;
+using Hangfire.Common;
+
+namespace Hangfire
+{
+    public interface IRecurringJobManager
+    {
+        void AddOrUpdate(string recurringJobId, Job job, string cronExpression);
+
+        void AddOrUpdate(string recurringJobId, Job job, string cronExpression, TimeZoneInfo timeZone);
+
+        void AddOrUpdate(string recurringJobId, Job job, string cronExpression, TimeZoneInfo timeZone, string queue);
+
+        void Trigger(string recurringJobId);
+
+        void RemoveIfExists(string recurringJobId);
+    }
+}
--- a/src/Hangfire.Core/IRecurringJobManager.cs
+++ b/src/Hangfire.Core/IRecurringJobManager.cs
@@ -1,5 +1,4 @@
-﻿using System;
-using Hangfire.Common;
+﻿using Hangfire.Common;
 
 namespace Hangfire
 {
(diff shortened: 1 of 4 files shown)


src/Hangfire.Core/IRecurringJobManager.cs
```diff
@@ -0,0 +1,18 @@
...
+
+namespace Hangfire
+{
+    public interface IRecurringJobManager
+    {
+        void AddOrUpdate(string recurringJobId, Job job, string cronExpression);
+
+        void AddOrUpdate(string recurringJobId, Job job, string cronExpression, TimeZoneInfo timeZone);
+
+        void AddOrUpdate(string recurringJobId, Job job, string cronExpression, TimeZoneInfo timeZone, string queue);
```
odinserj: Other overloads are just a syntax sugar for this method, and shouldn't be included into the interface. Plus, `timeZone` and `queue` parameters were added since the initial release of this class. Therefore I'm expecting that there will be some other parameters as well. 

Since we can't modify interface methods without introducing breaking changes, I think there should be the `RecurringJobOptions` class that will contain `TimeZone` and `Queue` properties (and other after some time), and the final signature that should go to the interface will be like this:

```sharp
void AddOrUpdate(string recurringJobId, Job job, string cronExpression, RecurringJobOptions options);
```

arnoldasgudas: Thanks for the comment. I've added `RecurringJobOptions` class and updated `IRecurringJobManager` interface.

src/Hangfire.Core/IRecurringJobManager.cs
```diff
@@ -0,0 +1,15 @@
+﻿using Hangfire.Common;
+
+namespace Hangfire
+{
+    public interface IRecurringJobManager
+    {
+        void AddOrUpdate(string recurringJobId, Job job, string cronExpression);
```
odinserj: This method shouldn't be placed to an interface. This is an overload that may be implemented as an extension method that calls the method above with the default options. This will result in a much simpler interface.

src/Hangfire.Core/RecurringJobManager.cs
```diff
@@ -62,6 +62,17 @@ public RecurringJobManager([NotNull] JobStorage storage, [NotNull] IBackgroundJo
         }
 
         public void AddOrUpdate(
```
odinserj: This method should contain the code, not call another overload. Consider we added just another property to the `RecurringJobOptions` class, what overload to call in this case? All other methods are just overloads with some default parameters.

odinserj:
I've merged this PR into the `dev` branch, please see commits 141cbb3e4f1e84328874062a32271c18374468d5 and 911525287174b1ec69dda33108bbe2f7333c8c9c. Also I've made the following changes. Thank you for the PR!

* Remove overloads of the `AddOrUpdate` method.
* A couple of methods are extensions now to be able to use them from
`IRecurringJobManager`.
* Add null checks to `RecurringJobOptions` class
* Add some unit tests

PR #502, opened 2016-01-20
Requeue to initial queue

gandarez:
Fixed the problem when requeueing to default queue.

https://discuss.hangfire.io/t/queue-name-will-become-default-after-automatic-fail-retry/1268
https://discuss.hangfire.io/t/rescheduling-the-job-to-default-queue-despite-it-is-being-set-in-another-queue/1175/4


seertenedos:
Any idea when this fix will be merged in?

gandarez:
Still waiting priority

odinserj:
This is by design. The `Queue` property is related to the `EnqueuedState`, and not to background job itself. In Hangfire, background jobs don't know anything about queues, schedules, cron strings, etc. They only contain invocation data (type, method, arguments), have unique identifier and current state.

The current state determines the further processing logic, and may have an additional information, such as queue for enqueued jobs. Current state is volatile, and states are unrelated to each other. If you want to assign some parameter to a background job, you can use `SetJobParameter` and `GetJobParameter` methods of the `JobStorageConnection` class using job filters. 

This distinction of properties between jobs and states allows Hangfire to be highly extensible, allowing to add more features as plugins, including new states, new processing rules and so on. The feature you are proposing may be implemented using the following job filter.

```csharp
public class PreserveOriginalQueueAttribute : JobFilterAttribute, IApplyStateFilter
{
    {
        var enqueuedState = context.NewState as EnqueuedState;

        // Activating only when enqueueing a background job
        if (enqueuedState != null)
        {
            // Checking if an original queue is already set
            var originalQueue = JobHelper.FromJson<string>(context.Connection.GetJobParameter(
                context.BackgroundJob.Id,
                "OriginalQueue"));

            if (originalQueue != null)
            {
                // Override any other queue value that is currently set (by other filters, for example)
                enqueuedState.Queue = originalQueue;
            }
            else
            {
                // Queueing for the first time, we should set the original queue
                context.Connection.SetJobParameter(
                    context.BackgroundJob.Id,
                    "OriginalQueue",
                    JobHelper.ToJson(enqueuedState.Queue));    
            }
        }
    }

    {
    }
}
```

You can decorate a method using this filter, add this filter globally using `UseFilter` method (`GlobalConfiguration.Configuration`), or combine it with the `QueueAttribute` filter, it will work fine. I don't want to add as the default filter, because it requires additional storage round-trips. However, it should be documented as well, as an example.

gandarez:
@odinserj thanks for your reply.

I did few modifications on your example because I need it much dynamic since I do not have the queue name at that moment. GetStateData method is a new overload method based on its parent and returns the latest result from State table which guarantees that the queue will be set correctly.

```C#
public class PreserveOriginalQueueAttribute : JobFilterAttribute, IApplyStateFilter
{
    {
        var enqueuedState = context.NewState as EnqueuedState;
        
        if (enqueuedState != null)
        {
            var stateData = context.Connection.GetStateData(context.BackgroundJob.Id, EnqueuedState.StateName);

            if (stateData != null)
                enqueuedState.Queue = stateData.Data["Queue"];
        }
    }

    {
    }
}

public override StateData GetStateData(string jobId, string stateName)
{
    if (jobId == null) throw new ArgumentNullException("jobId");

    string sql = string.Format(@"
select Name, Reason, Data
from [{0}].State
where JobId = @jobId
and Name = @stateName", _storage.GetSchemaName());

    return _storage.UseConnection(connection =>
    {
        var sqlState = connection.Query<SqlState>(sql, new { jobId = jobId, stateName = stateName }).LastOrDefault();
        if (sqlState == null)
        {
            return null;
        }

        var data = new Dictionary<string, string>(
            JobHelper.FromJson<Dictionary<string, string>>(sqlState.Data),
            StringComparer.OrdinalIgnoreCase);

        return new StateData
        {
            Name = sqlState.Name,
            Reason = sqlState.Reason,
            Data = data
        };
    });
}
```

gandarez:
@odinserj I followed your example and it seems not to be working.

image

odinserj:
@gandarez, how do you enqueue your background job, and how do you apply the filter? I see the `OriginalQueue` parameter is set to *default*, this means filter was plugged in, but it hasn't applied, or rewritten previous value for some reason.

gandarez:
@odinserj The job listed above is enqueded exactly the same way as recurring one. The difference is that I created a new class (inherited from recurring) that internally is called as persistent because I need to trigger through the Dashboard all the jobs non-recurring. So following this scenario I applied the filter you wrote:
```C#
.UseFilter(new PreserveOriginalQueueAttribute());
```
When I run the Sample project from Hangfire solution it goes like a charm but if I do it from my windows service as explained above it simply doesn't apply the filter. 

Any idea?

odinserj:
Hi @gandarez, is the problem with `PreserveOriginalQueueAttribute` still relevant?

gandarez:
@odinserj I should test it again. I hope do it soon and tell you.

gandarez:
@odinserj It's not working when the recurring job has been triggered manually.

Triggered by recurring job scheduler
image

Triggered using recurring job manager
image

gandarez:
@odinserj help me!

ZA1:
I needed this feature and the filter that @odinserj provided works. We have a separate service that runs the tasks and the dashboard is just a view into that database. What got me is that I did not add the filter to the Dashboard configuration so when re-queuing the jobs form the dashboard the job landed up in the default queue.

After I added the filter to the dashboard queue as well as the service configuration it worked like a charm.

Zapnologica:
Is there any update on this?

We are having endless issues with jobs requiring on the wrong queue.  This is very clearly a bug.

What is the point of even having an Queue attribute when it SOMETIMES looks at it?  The whole functionality of hangfire is to have retry logic.  So this honest just seems silly.  IF you are going to add a feature it must work properly.

I have seen a number of people complaining about this online?  Is there any known solution? Or way in which we can fix this?

lennartquerter:
I Second @Zapnologica , We are facing the same issue. We would really like to use the Queues for different services connected to one Hangfire database as it makes alot of sense in our environment.

berdem:
> I Second @Zapnologica , We are facing the same issue. We would really like to use the Queues for different services connected to one Hangfire database as it makes alot of sense in our environment.

Does this help you?

https://earwicker.com/Hangfire_A_Tale_of_Several_Queues.html

oguzhankahyaoglu:
bump this thread and fix it please :/

taylor-viventium:
Still a problem in 2025. This should be standard functionality.

Carben-dev:
> Still a problem in 2025. This should be standard functionality.

I totally agree. When a Recurring Job is retried, it ends up in a different queue. This caused some issues in my codebase, but I never really thought about the queue aspect until I looked into it more closely. That’s when I realized that keeping the job in its original queue during requeueing isn’t the default behavior.

Issue #505, opened 2016-01-25
Add Arguments to Job description in Recurring Job page


yschiller:
/recurring displays only the Type.Method for the job, ignoring the parameters which are often critical in determining which job is which.

can you please add the arguments to the Job description? or perhaps add a job description field that can be used in RecurringJob.AddOrUpdate

xrkolovos:
why this is never discussed?

dcorbett-fivebridgesllc:
I don't know if this works for recurring jobs -- I'm not using them yet... but for regular jobs:

> 
        [Queue(BatchQueueNames.BatchCreateAnalysis), LogStatus, DisplayName("Create Analysis {0} {1} {2}")]
        void CreateAnalysisBatch(string batch, string section, string analysisRunId, AnalysisBatchRequest analysisBatchRequest, PerformContext performContext, IJobCancellationToken cancellationToken);

works quite well so that the first 3 parameters are displayed in the Hangfire UI, at least until an error happens and that info is replaced in the list.

xrkolovos:
Nice I ll try it.

MovGP0:
While this may work, I don't like it. There should be the possibility to declare the display name when scheduling the job imperatively: 

```csharp
var state = new EnqueuedState(queue: "Default", displayName: $"Processing order for {customerName}");
var jobId = BackgroundJobClient.Create<ISomeCustomerJob>(client => client.ExecuteAsync(order, null, JobCancellationToken.Null), state);
```

This way I don't need add additional parameters that is not relevant for the execution of the job (ie. the description needs the customer name, but the job only needs the customer ID), because otherwise it would be a violation of the [Law of Demeter](https://en.wikipedia.org/wiki/Law_of_Demeter).

PR #507, opened 2016-01-25
https://github.com/HangfireIO/Hangfire/issues/505 - Added Args to Job

yschiller:
Added Args to the Job Tostring so it is displayed with Arguments in the dashboard


odinserj:
Arguments may add a lot of noise, especially when a user passes a complex object. Instead, you can use the `DisplayNameAttribute` for methods you want:

```csharp
[DisplayName("MyClass.MyMethod({0}, {1})")]
public void MyMethod(string arg1, int arg2)
```

yschiller:
Thanks great. Thank you!

PR #519, opened 2016-02-25
Fix for ToGenericTypeString method to support nested classes

patrykpiotrmarek:
Current version of ToGenericTypeString method doesn't support nested classes. 
This is my fix for it. Commit also contains some tests to prove its correctness.


odinserj:
Will be happy to merge this PR, but please target it against the `dev` branch. Special thanks for unit tests, much simpler to see the changes :-)

PR #520, opened 2016-02-26
Zero latency processing

bgurmendi:


gandarez:
This looks good!

burningice2866:
This is very clever, i'll implement it in my own Storage Provider for sure.

odinserj:
Small, but great optimization, especially for development environments. Thanks, @bgurmendi, merging!

odinserj:
I've changed implementation a bit in the 01897981381f0b97bdf5768156daf272095cb834 commit. Now the event is set only when the `AddToQueue` method is called.

ColinBlair:
What happens if there is already an ambient transaction? I am guessing that the contents of _afterCommitCommandQueue are then being executed before the ambient transaction completes unless there is a volatile resource manager involved somewhere.

Zoxive:
Can anyone confirm this is working as intended in 1.6.8?

With some simple diagnostics using it appears like the NewItemInQueueEvent is firing, but there is still a few second delay before my task is executed.

I added the following with a Task.Run() in my Startup.cs 

```csharp
var type = Type.GetType("Hangfire.SqlServer.SqlServerJobQueue, Hangfire.SqlServer");

System.Diagnostics.Trace.WriteLine("Found Hangfire Object");
System.Diagnostics.Trace.WriteLine(type != null? type.FullName : "(null)");

foreach (var p in type.GetFields(BindingFlags.Static | BindingFlags.NonPublic))
{
    if (p.Name == "NewItemInQueueEvent")
    {
        System.Diagnostics.Trace.WriteLine("Found Hangfire Property");
        resetEvent = (AutoResetEvent)p.GetValue(null);
    }
}

while (true)
{
    WaitHandle.WaitAny(new[] { resetEvent }, TimeSpan.FromMinutes(15));

    System.Diagnostics.Trace.WriteLine("GOT EVENT!!@");
}
```

As soon as i enqueue my background job i see "GOT EVENT!!@" in my trace, but i still have a multi second delay before my job even starts executing.

Zoxive:
I think i figured it out.

If you have more than one app.UseHangfireServer() it doesnt work.

I currently have 3 so i can set up different filters.

PR #517, opened 2016-02-20
Implementing multilanguage support in Dashboard.

Elph:
Hi, I was working a little in this feature and I hope this Pull Request could be useful to other people.
Not sure about the Nuget implementation, so I've put everithing on the Dashboard Content folder

Trying to implement #331 

* The language is defined in web.config.
* Extracted all resources from *.cshtml
* Added moment-with-locales javascript and initialized in hangfire.js, so all the time texts also are translated. I guess moment.min.js could be removed completely
* I Created a Single resx file and tried to reuse most of the texts.

Let me know if I can improve anything else.

Issue #331, opened 2015-03-26
Support other languages in Dashboard


bhugot:
Hello,

Would be great if the dash board could be display in others langages.

Thanks.

odinserj:
Yep, it would be awesome. There are two sub-tasks: 

1. Extract resources.
2. Think about [localization options](http://docs.nuget.org/Create/Creating-Localized-Packages).

I'll be glad to accept a pull request that implement this, after discussing localization approaches (under item 2).

odinserj:
Is anybody wants to do this?

schulz3000:
I'm interested in doing point 1. Can I start with it?
Also I think we need a third point here: Ability of setting the language. Should it be autodetected by browserlanguage or can the user set it manually?

odinserj:
I think we should not reinvent the wheel here and use regular resource files to store translations and use regular .NET methods to retrieve the localized strings that use `CultureInfo.CurrentUICulture`. The main point is how to deliver the resource strings – use additional NuGet packages for satellite assemblies or store translations in one NuGet package.

sackerud:
Any movement on this issue? I've done a couple of translations before and contribute with swedish. I'm gravitating towards the single package approach as described on nuget.org.

gandarez:
I can help on translating to Portuguese as well.

Elph:
I created a PR with this feature, please let me know what do you think...

gandarez:
Looks good and as soon it gets approved by @odinserj I'll start translating to pt-BR.

Elph:
I can work on the spanish translation also when accepted.

gandarez:
@odinserj please tell me as soon as you get it accepted.

odinserj:
I've just merged the PR by @Elph. So nothing prevents us from adding translations :tada: We can start from embedding localized resources into Hangfire itself without spawning additional NuGet packages due to the simplicity of this approach.

PR #521, opened 2016-02-26
Implementing multilanguage support in Dashboard.

Elph:
Replaces PR #517, now pointing to the right /dev branch

Trying to implement #331 

* The language is defined in web.config.
* Extracted all resources from *.cshtml
* Added moment-with-locales javascript and initialized in hangfire.js, so all the time texts also are translated. I guess moment.min.js could be removed completely
* I Created a Single resx file and tried to reuse most of the texts.

Let me know if I can improve anything else.


odinserj:
@Elph, I've just merged the PR. Thank you a lot for the work!

KveRa:
@Elph Is it possible, (as it is now) to use a custom resource in the project using Hangfire to add a custom Translation? If yes, how ?

burningice2866:
@KveRa Yes you can. The easiest way is to fork the repository and copy https://github.com/HangfireIO/Hangfire/blob/master/src/Hangfire.Core/Dashboard/Content/resx/Strings.resx into `Strings.<language>.resx` and start translating. When you're finished you compile the project which will then produce a `<langage>` folder in the bin-output which you can copy to the bin-folder of your project, and voila, you now have everything translated into your language.

When you're done it would be a nice gesture to do a pull-request to get the language added to core.

PR #524, opened 2016-03-03
Refactored SqlServerStorage to use the more general DbConnection

tystol:
I use HangFire and EntityFramework together with a shared connection, allowing for a shared transaction (without escalation to DTC):

```
builder.Register(c =>
{
    var connection = c.Resolve<ConnectionManager>().Connection;
    return new BackgroundJobClient(new SqlServerStorage((SqlConnection) connection));
}).As<IBackgroundJobClient>();
```

As you can see, I need to cast the DbConnection to a SqlConnection. This was working fine, until I installed Glimpse. They wrap all DbConnections with their own decorator class, which derives from DbConnection, but not SqlConnection (which makes sense). This then causes this cast to fail, so right now I am doing the following:

```
builder.Register(c =>
{
    var connection = c.Resolve<ConnectionManager>().Connection;
    var sqlConnection = connection as SqlConnection;
    if (sqlConnection == null)
    {
        var glimpseConnection = connection as GlimpseDbConnection;
        if ( glimpseConnection != null )
            sqlConnection = glimpseConnection.InnerConnection as SqlConnection;
    }
    return new BackgroundJobClient(new SqlServerStorage(sqlConnection));
}).As<IBackgroundJobClient>();
```

It works, but now the Hangfire connection is not profiled.

This PR simply changed all SqlConnection references to DbConnections, and it all seems to work fine. I've built my application against this and also tested there, again working fine.

Thoughts on pulling this in? Was their a specific reason for using SqlConnection?


burningice2866:
+1

odinserj:
@tystol, please make this PR target to the `dev` branch, I'll release this with 1.6.

PR #526, opened 2016-03-08
Add support for async/Task methods

### LINKED ISSUE #150: Support for `async` background job methods
Heavy I/O bound background jobs consume worker thread inefficiently – instead of doing the real work they wait for the completion of I/O operation(s). Currently if an application has a lot of such background jobs, it is better to increase the worker pool size to process jobs more efficiently. However increased worker pool causes performance problems for CPU-bound jobs.

Instead of waiting for the I/O completion, it would be useful to yield the worker thread to process another background job instead, for example, using the default mechanism with `async` and `await` keywords:

``` csharp
public async Task IOBoundMethod()
{
    var documents = await _documentsService.FetchAll();
    documents.Process();
}
```

But we should ensure that the completion is being called inside the _Hangfire's worker thread_, and not in _CLR's thread pool thread_ (to not to compete with request processing pipeline, use additional storage connections and so on). Another consideration should be dedicated to job filters – `CaptureCultureAttribute` changes the thread context, and it should be changed again on each continuation. There are also problems with reliability (outstanding jobs should be requeued), shutdown events (outstanding operations should be canceled), and perhaps with something else.

tuespetre:
Fixes #150

I also added a test that confirms that a method can accept a standard .NET `CancellationToken`.


burningice2866:
Thats was easy... is that really it? :)

odinserj:
Yep, that is it :smile: However, I'd like to add the following stuff after merging this. But before, could you please open a new PR that targets the `dev` branch? I want to incorporate this change into the 1.6 release.

* Prohibit the use of `async void` methods by throwing an exception when checking a method type.
* Pass `InnerException` instead of `AggregateException` itself to a `JobPerformanceException` instance, if there is only one inner exception, leaving aggregate exception as-is, when there are additional exceptions.
* Add a special case for `TaskCanceledException`, when passing a shutdown token to an awaitable operation. Currently it causes background job to fail on a server shutdown, but I'm expecting it will be silently placed back to a queue.
* Disable JSON serialization for arguments of the `CancellationToken` type, it is completely unnecessary.

Warning 4014 really annoying and pointless in the case of expression trees, since they aren't lambdas.

tuespetre:
@odinserj will do, thanks

PR #533, opened 2016-03-16
Updating StackTraceParser to prevent critical flaw

lukerogers:
The StackTraceParser had a critical flaw that caused the server to hang when specific error messages were returned (https://github.com/atifaziz/StackTraceParser/issues/4). This just replaces the parser with the updated version to avoid such problems.


odinserj:
@lukerogers, thank you for the PR!

Issue #500, opened 2016-01-19
BackgroundJob.Delete may hang forever


jods4:
If you pass a non-existing id to the public API `BackgroundJob.Delete`, it will loop forever.

The bug is in `BackgroundJobStateChanger.GetJobData()` look here:
https://github.com/HangfireIO/Hangfire/blob/master/src/Hangfire.Core/States/BackgroundJobStateChanger.cs#L130

This method will never stop trying to find a job even if it doesn't exist.

If you look at this comment in caller code, this is clearly not the intended behaviour:
https://github.com/HangfireIO/Hangfire/blob/master/src/Hangfire.Core/States/BackgroundJobStateChanger.cs#L59-L63

gandarez:
I suppose there's a loop to wait until the the data gets released (unlocked). But I think it might try only twice instead of an infinite loop!

jods4:
@gandarez  I'm not even sure this is the case!

If you look at what `SqlServerConnection.GetJobData()` does under the hood, it comes down to [simplified] this:
```csharp
var jobData = connection.Query<SqlJob>("select * from Job where id = @id", new { id = id }).SingleOrDefault(); 
if (jobData == null) return null; 
```

There is no locking or whatever that justifies waiting here. We plainly look if the job is in the table or not.

Now, I don't fully grasp the Hangfire codebase, maybe those considerations apply to other providers?

captain828:
This is an actual issue I am having right now with the latest Hangfire version (1.5.3) after we migrated our Hangfire database.

It just hangs while attempting to retrieve a job that does not exist and the culprit does seem to be the BackgroundJobStateChanger.GetJobData() method.

@odinserj - please advise.

danoleary:
I ran into this issue yesterday, in 1.5.3. Could BackgroundJobStateChanger.GetJobData() just return null if no job is found for a given id?

captain828:
https://github.com/HangfireIO/Hangfire/blob/master/src/Hangfire.Core/States/BackgroundJobStateChanger.cs#L59-L63

So the null case should be covered already.

IMHO the https://github.com/HangfireIO/Hangfire/blob/master/src/Hangfire.Core/BackgroundJobServerOptions.cs#L76 setting or a new setting should be used to set the time-out for getting job data.

Thoughts?

danoleary:
[...]

Issue #294, opened 2015-01-28
Enqueued job is not being processed with MSMQ or RabbitMQ


odinserj:
From forum topic, please see http://discuss.hangfire.io/t/jobs-being-created-in-database-stuck-as-enqueued/583 for the details.

> While performing a simple Enqueue such as:
> BackgroundJob.Enqueue(() => Console.WriteLine("Test!"));
> Hangfire will only attempt to process the job around 60% of the time. Further research shows that the job is being successfully inserted into the HangFire.Job table, but it is never processed.

odinserj:
The problem was related to a very quick worker that picked up a background job that was not fully initialized for the enqueued state. I've fixed it by waiting for a full initialization (with timeout) before trying to change the state to Processing. If timeout expires (or when process is killed), the job simply moves back to the queue.

PR #535, opened 2016-03-17
Fix for #500, BackgroundJob.Delete may hang forever

danoleary:
Added null handling to prevent an infinite loop if an attempt is made to delete a job
that doesn't exist

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -2,7 +2,7 @@
 # see travis-ci.org for details
 
 language: csharp
-sudo: false
+sudo: required
 
 # Make sure build dependencies are installed.
 install:
--- a/src/Hangfire.Core/States/BackgroundJobStateChanger.cs
+++ b/src/Hangfire.Core/States/BackgroundJobStateChanger.cs
@@ -127,6 +127,12 @@ namespace Hangfire.States
             while (true)
             {
                 var jobData = context.Connection.GetJobData(context.BackgroundJobId);
+
+                if (jobData == null)
+                {
+                    return null;
+                }
+
                 if (jobData != null && !String.IsNullOrEmpty(jobData.State))
                 {
                     return jobData;
--- a/src/Hangfire.Core/States/BackgroundJobStateChanger.cs
+++ b/src/Hangfire.Core/States/BackgroundJobStateChanger.cs
@@ -122,31 +122,13 @@ namespace Hangfire.States
 
         private static JobData GetJobData(StateChangeContext context)
         {
-            var firstAttempt = true;
-
-            while (true)
-            {
-                var jobData = context.Connection.GetJobData(context.BackgroundJobId);
-
-                if (jobData == null)
-                {
-                    return null;
-                }
-
-                if (jobData != null && !String.IsNullOrEmpty(jobData.State))
-                {
-                    return jobData;
-                }
-
-                if (context.CancellationToken.IsCancellationRequested ||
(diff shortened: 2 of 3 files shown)


captain828:
Is the loop even needed any more?
Also, what is the scenario for jobs with no state?

danoleary:
With the change I have made, the loop will only come into play when context.Connection.GetJobData(context.BackgroundJobId) returns a job with a null or empty state. 

Given I haven't been able to find any mention of issues caused by empty states causing Hangfire to hang, should we leave this behaviour as is?

captain828:
Since I don't know the full implementation of the cancellation token, I can't tell if this is going to break it or not.

As far as I can understand, a job is always in a given state, as such:
```
if (jobData == null)
{
      return null;
}

if (jobData != null && !String.IsNullOrEmpty(jobData.State))
{
      return jobData;
}
```
will guarantee an exit, which in turn means that:
```
if (context.CancellationToken.IsCancellationRequested ||
    context.CancellationToken == CancellationToken.None)
{
       return null;
}
```
will never be executed, potentially breaking the cancellation token feature.

danoleary:
If a job is guaranteed to have a state, could we just simplify the method to this?
````
private static JobData GetJobData(StateChangeContext context)
{
   if (context.CancellationToken.IsCancellationRequested ||
                    context.CancellationToken == CancellationToken.None)
   {
       return null;
   }

   return context.Connection.GetJobData(context.BackgroundJobId);
   }
````

captain828:
Yeah, that should work just fine.

captain828:
Looks like the RabbitMQ storage provider was not very happy with this :)

danoleary:
It does! I've ran all of the tests successfully locally though, so I'm not really sure how to troubleshoot this. Do you know if there's a way of kicking off another build without making changes?

captain828:
Yeah, looks like the build errored, not failed.

Close the PR and then re-open it. This should force a new build.

captain828:
Same issue:
`Test execution time exceeded: 30000ms`

@odinserj - could you have a quick look to see if the RabbitMQ server is reachable?

danoleary:
Thanks for looking into this, shall we leave this pull request as it is then?

captain828:
Could you kindly change line 5 in https://github.com/HangfireIO/Hangfire/blob/master/.travis.yml to `sudo: required` and commit it?

If it still errors out, then there really isn't much else we can do.

captain828:
There we go! :+1:

odinserj:
This issue is already resolved in PR #469. I've just released [1.5.4](https://github.com/HangfireIO/Hangfire/releases/tag/v1.5.4) that fixes the problem.

odinserj:
Infinite loop is still required, because some queue providers, such as Hangfire.SqlServer.MSMQ and Hangfire.SqlServer.RabbitMQ may insert a background job identifier to a queue (and make it available for workers) *before* background job is saved to a job storage. 

This may happen, because background job creation is held within a transaction, and queue provider may insert a job before that transaction is committed. And this results into issue described in #294.

captain828:
I knew it was there for a reason :)

Thanks for looking into this and for pushing a new version!

odinserj:
Yep, but I'm ashamed to release the fix several months after it was committed to the master :frowning:

captain828:
Well you definitely shouldn't feel ashamed that you created such an amazing and easy-to-use scheduler!

PR #540, opened 2016-03-24
Add support for async Task methods

### LINKED ISSUE #150: Support for `async` background job methods

tuespetre:
Fixes #150

I also added a test that confirms that a method can accept a standard .NET `CancellationToken` and addressed two of the items mentioned on the previous PR (disallowing `async void` methods and unwrapping `AggregateException`s when possible.)

### DIFF
--- a/src/Hangfire.Core/Common/Job.cs
+++ b/src/Hangfire.Core/Common/Job.cs
@@ -21,6 +21,7 @@ using System.Linq.Expressions;
 using System.Reflection;
 using System.Threading.Tasks;
 using Hangfire.Annotations;
+using System.Runtime.CompilerServices;
 
 namespace Hangfire.Common
 {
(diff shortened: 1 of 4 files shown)


odinserj:
Thanks @tuespetre, merged! I'll release it with 1.6.0-beta1 next week.

tuespetre:
Now that I'm starting to use `1.6.0-beta1` it's occurring to me that although these changes would let one use `async` and other `Task`-returning methods, the 'core' of Hangfire still will not *truly* process them in an async manner. What are your thoughts on this?

odinserj:
Yep, this is just a syntax sugar, I've already [commented this](https://github.com/HangfireIO/Hangfire/issues/150#issuecomment-201722775). Real asynchrony will be a very hard-to-implement feature for Hangfire.

tuespetre:
Kewl beanz. I must have missed your comments there. Thanks!

PR #555, opened 2016-04-15
Enable backtracking from background job to recurrent job

### LINKED ISSUE #554: Unable to get RecurrentJobID associated with BackgroundJob
As it stands now the context of a BackgroundJob that is started from a RecurrentJob does not know which RecurrentJob it came from. I propose to add support for this kind of backtracking.

One way I need to use this feature is to disable job concurrency for individual RecurrentJobIDs similarly to what was proposed here https://discuss.hangfire.io/t/job-reentrancy-avoidance-proposal/607

teodimache:
Enable the Background Job context to hold the Recurrent Job ID that spawned it.
Fixes #554


teodimache:
Couldn't decide between using either Items or Parameters to hold the RecurrentJobId so i added both.

odinserj:
I think "Parameters" would fit better, because the whole background job is based on the recurring job. Items collection is temporary, for sharing data between filters. Would you like to remove it from the `Items`? I will be happy to merge this PR.

teodimache:
Make sense, I removed it from Items. Thanks!

odinserj:
Thanks, will merge the PR soon to the `dev` branch :tada:. Now we may also link background job with its recurring job in the dashboard.

odinserj:
@teodimache, I've merged this to the `dev` branch. Changes will be available with the 1.6.0-beta3 release. Thanks for the PR!

teodimache:
Awsome! Looking forward to 1.6 :)

PR #561, opened 2016-04-22
Add queue name validation to the RecurringJobScheduler class

pzielinski86:
I moved logic of validation to QueueValidator. It should fix the issue when adding a job with an invalid name is possible, but triggering it always fails without any exception.


pzielinski86:
Any thoughts?

odinserj:
@pzielinski86, what is the reason of putting the logic into a separate public class with a single method?

odinserj:
However, I like to see the queue name check in the `RecurringJobManager` class :heart:

pzielinski86:
@odinserj:
I didn't want to duplicate the validation logic in 2 places. Maybe there is a better place, like an existing namespace\class?

What do you think?

odinserj:
We can just change the visibility of the method from `private` to `internal` or `public` without adding new entities (as they add unnecessary complexity in this case). I don't see any advantages of making it public, but leaving this decision to you!

pzielinski86:
I think you are right. EnqueuedState is used in RecuringJobManager anyway so I will make it internal static in EnqueuedState.

odinserj:
Ok, will wait for the updated PR!

pzielinski86:
I marked the method as static and removed the created before class.

odinserj:
Thanks, @pzielinski86! I've just merged this PR. Will be released with 1.5.7

Issue #567, opened 2016-05-12
Recurring Jobs not Running Correctly when summer time ends in BST


DownChapel:
After the recent bug with time zones when moving into DST (https://github.com/HangfireIO/Hangfire/issues/529), I have been testing Hangfire v1.5.6 moving out of DST.

I am testing in British Standard Time, where on the last Sunday of October, the time at 2am changes back to 1am. When I test this in Hangfire, my jobs that run every minute, do not run for an hour when the clocks go back from 2am to 1am. However, they do start up again after that hour.

I have narrowed it down to the code in `ScheduleInstant`: https://github.com/HangfireIO/Hangfire/blob/v1.5.6/src/Hangfire.Core/Server/ScheduleInstant.cs#L46-L55

The following test code using 1:59 in BST (0:59 in UTC) should return 1:00 in BST (1:00 UTC). However, it returns 2:00 in BST (2:00 UTC). This is because `NCrontab` doesn't deal with time zones, so if you pass 1:59. you will get 2:00am back.

```
var gmtStandardTime = "GMT Standard Time";
var gmtStandardTimeZone = TimeZoneInfo.FindSystemTimeZoneById(gmtStandardTime);

var crontabSchedule = CrontabSchedule.Parse("* * * * *");

// 0:59 in UTC = 1:59 in BST. 1:00 UTC = 1:00 BST

var utcStartTime = new DateTime(2016, 10, 30, 0, 59, 0);
var startTimeInBst = TimeZoneInfo.ConvertTimeFromUtc(utcStartTime, gmtStandardTimeZone);
var endTimeInBst = TimeZoneInfo.ConvertTimeFromUtc(new DateTime(2016, 10, 30, 2, 20, 0), gmtStandardTimeZone);

var list = crontabSchedule
    .GetNextOccurrences(
        startTimeInBst,
        endTimeInBst)
    .ToList();
```

I'm happy to try and fix this, but I'm not sure how to go about it. Any tips would be appreciated!

odinserj:
@DownChapel, thanks for reporting this! Indeed, NCrontab doesn't handle daylight savings at all, and works correctly only for UTC timezone as it uses simple second/minute/whatever increment in its [`GetNextOccurrence`](https://github.com/atifaziz/NCrontab/blob/master/NCrontab/CrontabSchedule.cs#L212) method, without the knowledge about timezones and daylight savings. As a result – we have missed values when the clock is moved backwards.

[...]

PR #568, opened 2016-05-12
Changing ScheduleInstant to use UTC to get next run date

DownChapel:
Found an issue when the clocks go back jobs don't run for an hour. Reference: https://github.com/HangfireIO/Hangfire/issues/567

ScheduleInstant uses NCrontab to get the next occurrence in the schedule. This doesn't support time zones so needs to be passed UTC.

### DIFF
--- a/src/Hangfire.Core/Server/ScheduleInstant.cs
+++ b/src/Hangfire.Core/Server/ScheduleInstant.cs
@@ -43,17 +43,7 @@ namespace Hangfire.Server
 
             NowInstant = nowInstant.AddSeconds(-nowInstant.Second);
 
-            var nextOccurrences = _schedule.GetNextOccurrences(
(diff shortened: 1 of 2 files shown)


DownChapel:
I may well be missing something, but seeing as `ScheduleInstant` stores dates in UTC, then calculating the next occurrence should be done in UTC too. Otherwise you get issues with time zones like the test (`NextInstant_DoesntThrow_NearEndOfDaylightSavings`) I've added to `ScheduleInstantFacts` which fails with the existing implementation .

odinserj:
@DownChapel, I've [commented](https://github.com/HangfireIO/Hangfire/issues/567#issuecomment-219024512) the original issue. Closing the PR, let's discuss the implementation for the fix first.

DownChapel:
Cool, thanks

PR #570, opened 2016-05-13
passing params in job creation phase

abdelrady:
We needed a way to pass parameters when we create recurring jobs.

We don't need to pass those parameters to every job we declare as this is irrelevant to the job task/business.
We were using the JobFilterAttribute to get params from IOC container and adding params in createcontext and receiving them at job execution phase, but this approach wasn't helping in case of recurring jobs because IOC wasn't able to get those param values from its old context or instance used to queue this job from the first place(our job queuing instance - when we called AddOrUpdate - destroyed on recurring jobs).

So we needed a way to attach params to recurring jobs in the queuing/serialization stage so we can get them again from Parameters list in the createcontext instance when job is loaded/deserialzed.

We tried to expose an access to Parameters property in CreateContext instance to let us add to it and it can be serialized/saved in disk when adding a recurring job.


apetkovic:
We have the exact same problem. Please add this to the release.

marcrocny:
Coming up on a year.  Thanks for your work on this. This may be a deal-breaker for us, would really like to be able to use in our solution.

MattRamsay:
+1 on this, would fine this extremely useful right now

burningice2866:
You might want to look at starting your pull-request over based on the current version of Hangfire

brian-knoll-micronetonline:
ill ask the person that opened it originally to close it. then ill open my own with fewer changed files

abdelrady:
Closing this PR and there should be another pull request with changes based on latest version of Hangfire.

PR #573, opened 2016-05-18
Add missing transaction to KeepAliveQuery.

maxhauser:
It looks like the `transaction` argument is missing from the `Execute` call in the `KeepAliveQuery` method.


odinserj:
OMG, how I could miss it. Thanks for the fix!

PR #574, opened 2016-05-18
Add spanish language for the Dashboard

Elph:
Adding Spanish :es: translation for Hangfire. Any correction for the strings translated will be great.

After translating I saw that all the Metrics titles always showed in english. It was because they were static, so the title was always on the default culture before setting any other culture. 
I modified them to use a TitleResourceKey and load the title from resources when asked.

### DIFF
--- a/src/Hangfire.Core/Dashboard/Content/js/hangfire.js
+++ b/src/Hangfire.Core/Dashboard/Content/js/hangfire.js
@@ -66,7 +66,8 @@
 
             var hoverDetail = new Rickshaw.Graph.HoverDetail({
                 graph: this._graph,
-                yFormatter: function (y) { return Math.floor(y); }
+                yFormatter: function (y) { return Math.floor(y); },
+                xFormatter: function (x) { return moment(new Date(x * 1000)).format("LLLL"); }
             });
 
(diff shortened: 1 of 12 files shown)


odinserj:
Yay, can't wait to merge it!

Regarding metrics, you are right, we need something to handle the titles. However, the proposed solution contains breaking changes, and can't be done in 1.x – some users may get compile-time errors due to this change. Is it possible to somehow leave the `Title` property and existing constructors untouched?

Elph:
Yeah, sure.. I'll do the changes when I've got some time. :+1:

Elph:
I did the proposed changes, keeping the DashboardMetric class untouched. Now the translation is done on the cshtml file instead. Still I use the title property to put the Name of the key that contains the resource, but if it's null it fallsback to the original string. I hope that is acceptable.

Also I added some changes to localize a couple of things I messed before.
* The Date on the graph now  is localized using moment.js
* The tooltip when doing hover on the graph also uses a localized string (succeeded, failed) obtained through data attributes

Let me know what do you think and if something else is needed

odinserj:
Awesome, @Elph! Merging this PR :ship:  :tada:

PR #575, opened 2016-05-19
Allows JobActivator resolve dependencies using DI Container

rafaelamaralsilva01:
We try to solve a problem on JobActivator, when we are using MEF (Microsoft Extensability Framework) to load plugin.dll which contains the action we want to schedule.

The way I understand, it is impossible to use Type.GetType("SomeTypeLoadFromPlugin") for week references.

Now, I made possible to use Scope to resolve by type name as string and my application I use Autofac with MEF to load my plugin.

### DIFF
--- a/src/Hangfire.Core/JobActivator.cs
+++ b/src/Hangfire.Core/JobActivator.cs
@@ -17,6 +17,8 @@
 using System;
 using System.Collections.Generic;
 using Hangfire.Annotations;
+using Newtonsoft.Json.Serialization;
+using Newtonsoft.Json.Converters;
 
 namespace Hangfire
 {
@@ -47,6 +49,12 @@ namespace Hangfire
             return Activator.CreateInstance(jobType);
         }
 
+        public virtual object ActivateJob(string jobType)
+        {
+            var type = Type.GetType(jobType);
+            return type != null ? this.ActivateJob(type) : null;
+        }        
+
         public virtual JobActivatorScope BeginScope()
         {
             return new SimpleJobActivatorScope(this);
@@ -66,6 +74,19 @@ namespace Hangfire
             public override object Resolve(Type type)
             {
                 var instance = _activator.ActivateJob(type);
+                var disposable = instance as IDisposable;
+
+                if (disposable != null)
+                {
+                    _disposables.Add(disposable);
+                }
+
+                return instance;
+            }
+
+            public override object Resolve(string typeName)
+            {
+                var instance = _activator.ActivateJob(typeName);
                 var disposable = instance as IDisposable;
 
                 if (disposable != null)
--- a/src/Hangfire.Core/Storage/InvocationData.cs
+++ b/src/Hangfire.Core/Storage/InvocationData.cs
@@ -46,7 +46,7 @@ namespace Hangfire.Storage
         {
             try
             {
-                // Use DI to solve dependencies.
+                // Use Dependency Injection to solve dependencies.
                 using (var scope = JobActivator.Current.BeginScope())
                 {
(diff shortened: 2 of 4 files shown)


odinserj:
Hi @rafaelamaralsilva01. Thanks for the PR, but it is very hard even to review it:

1. PR description doesn't contain enough details – I still don't understand why the current solution doesn't work for you.
2. [Diff](https://github.com/HangfireIO/Hangfire/pull/575/files) doesn't show modified lines, looks like line endings were changed. Review will take much more time, and line endings aren't consistent with other project files.
3. Looks like it contains breaking changes. I really don't like to bring them in 1.x version (trying hard to follow [SemVer](http://semver.org/) where possible), because they cause compile time and runtime errors and require other users to do a lot of work.

Without addressing this issues I can't even tell you when I'll finish the review phase and give you the required feedback.

src/Hangfire.Core/Storage/InvocationData.cs
```diff
@@ -1,238 +1,253 @@
...
+        public string Arguments { get; set; }
+
+        public Job Deserialize()
+        {
+            try
+            {
+                // Use Dependency Injection to solve dependencies.
+                using (var scope = JobActivator.Current.BeginScope())
+                {
+                    var type = scope.Resolve(Type)?.GetType() ?? System.Type.GetType(Type, throwOnError: true, ignoreCase: true);
```
rafaelamaralsilva01: Here I need to use my DI Container to resolve my type, because the type which I want to use here came from a plugin (Load by MEF, week reference).

src/Hangfire.Core/Storage/InvocationData.cs
```diff
@@ -1,238 +1,253 @@
...
+        public List<Type> GetParameters(JobActivatorScope scope)
+        {
+            List<Type> parameterTypes = new List<System.Type>();
+            var parameters = JsonConvert.DeserializeObject<string[]>(ParameterTypes);
+
+            foreach (var current in parameters)
+                parameterTypes.Add(scope.Resolve(current).GetType());
+
+            return parameterTypes;
+        }
```
rafaelamaralsilva01: Here I have the same problem, my parameter came from a Plugin load by MEF. Then to solve then I need to use my DI Container.

src/Hangfire.Core/JobActivator.cs
```diff
@@ -1,88 +1,109 @@
...
+        {
+            return Activator.CreateInstance(jobType);
+        }
+
+        public virtual object ActivateJob(string jobType)
+        {
+            var type = Type.GetType(jobType);
+            return type != null ? this.ActivateJob(type) : null;
+        }        
+
```
rafaelamaralsilva01: This method makes possible resolve a type base on string name.

src/Hangfire.Core/JobActivator.cs
```diff
@@ -1,88 +1,109 @@
...
+                if (disposable != null)
+                {
+                    _disposables.Add(disposable);
+                }
+
+                return instance;
+            }
+
+            public override object Resolve(string typeName)
```
rafaelamaralsilva01: Same here, only make possible use type name as string.

stajs:
FWIW, I have MEF and Hangfire working. My stack is:

- Hangfire 1.5.2
- DNX RC1, using the new built-in DI.

This MEF/dependency structure works:

- Starting process that loads other plugins.
  - Plugin, that has background job to run.

This doesn't work out of the box; I haven't investigated this structure to see why not and it might be what you are trying to address with the PR:

- Starting process that loads other plugins.
  - Plugin, that has dependency on another DLL.
    - DLL that the plugin depends on, that has background job to run.

Based on that, I suggest that the PR should have started out as an issue, and we could potentially work through the problem without changes to Hangfire.

PR #581, opened 2016-05-27
Fix dashboard crashing when using MSMQ

yangman:
This fixes the dashboard crashing when using MSMQ, as discussed in: https://discuss.hangfire.io/t/dashboard-appears-broken-in-hangfire-1-5-2/1438/12


odinserj:
@yangman, thanks a lot for fixing this! The fix will be released with upcoming 1.5.7. By the way, you can post an answer to numerous StackOverflow questions regarding getting the MSMQ queue size, none of them contain the full answer currently.

PR #539, opened 2016-03-24
Refactored SqlServerStorage to use the more general DbConnection - Take2

tystol:
Same as https://github.com/HangfireIO/Hangfire/pull/524, just targetting dev branch

PR #594, opened 2016-06-15
Sql Server abstraction

deathgore:
Had to use SqlServerStorage with a custom DbConnection, but SqlServerStorage is using specific class SqlConnection which is sealed.
   - SqlConnection  replace with IDbConnection
   - SqlTransaction replaced with IDbTransaction

Usage:
  .UseStorage(new SqlServerStorage(new CustomSqlConnection("Hangfire")));

### DIFF
--- a/src/Hangfire.SqlServer/SqlServerStorage.cs
+++ b/src/Hangfire.SqlServer/SqlServerStorage.cs
@@ -1,23 +1,24 @@
 // This file is part of Hangfire.
 // Copyright © 2013-2014 Sergey Odinokov.
-// 
+//
 // Hangfire is free software: you can redistribute it and/or modify
-// it under the terms of the GNU Lesser General Public License as 
-// published by the Free Software Foundation, either version 3 
+// it under the terms of the GNU Lesser General Public License as
+// published by the Free Software Foundation, either version 3
 // of the License, or any later version.
-// 
+//
 // Hangfire is distributed in the hope that it will be useful,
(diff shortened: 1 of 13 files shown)


src/Hangfire.SqlServer/SqlServerStorage.cs
```diff
@@ -180,22 +181,22 @@ internal void UseConnection([InstantHandle] Action<SqlConnection> action)
...
+        internal T UseConnection<T>([InstantHandle] Func<DbConnection, T> func)
         {
-            SqlConnection connection = null;
+            IDbConnection connection = null;
 
             try
             {
                 connection = CreateAndOpenConnection();
-                return func(connection);
+                return func((DbConnection)connection);
```
odinserj: Hacky. There are a lot of places where `IDbConnection` is enough, but actually, sooner or later, it will be casted to the `DbConnection`, giving us a run-time error.

odinserj:
Since we are required to use the `DbConnection` class (the `EnlistTransaction` method is declared there), we can't use more general `IDbConnection` interface in all the places, we can only dig down to the `DbConnection` class.

Support for `DbConnection` parameters were already added in the PR #539 and merged to the `dev` branch. It will be released with 1.6.0, and currently you can use [Hangfire 1.6.0-beta3](https://github.com/HangfireIO/Hangfire/releases/tag/v1.6.0-beta3) that is already available on NuGet.

PR #603, opened 2016-06-30
Drop constraints on Server.Id prior to altering it to avoid conflicts

ranieuwe:
Currently running schema 3, and the alter statements for [Hangfire].[Server].[Id] in the schema 4 => 5 transition table fail because of a the clustered index that is referencing the [Id].

This is solved by dropping the constraint, doing the alter, and adding the constraint back. 

Surprisingly enough I could not replay this with a default run of the (Default)Install.sql script. I can only replicate this with a copy of our production database. Manually recreating the table (drop + create to from SSMS) did not recreate this issue either.


odinserj:
Thanks for submitting the PR, @ranieuwe, merging!

PR #612, opened 2016-07-14
Notification System with Email Module

stevetsang:
Without having to open the Hangfire Dashboard to monitor the status of Hangfire Server it can be difficult to check their statuses.  We have created a notification system which is extendible by adding modules such as the email one we have created as an example.

The email notification module only works for **full .NET only** (does not work for .NET Core) and not sure if you want this email notification module moved to a different repo? If you want we could expand the email notification module to support .NET Core with the help of MailKit.

We have also updated the IStorageConnection interface with an extra method to implement dunno if you are happy with us adding this method or do you want us to move it out as part of a new interface?

Here is an example of setting it up in startup.cs :

```
List<string> toEmails = new List<string>() { "email1@testdomain.com" };
List<string> toEmails2 = new List<string>() { "email2@testdomain.com" };

// You can register custom configuration and email clients in the EmailNotification project. This example uses Sendgrid.
SmtpConfiguration sendgridConfig = new SmtpConfiguration("smtp.sendgrid.net", "username", "password", 587);
SmtpClientNotifier client =  new SmtpClientNotifier("fromemail@testdomain.com", "fromname", sendgridConfig);

//Only 2 types of events for now, it is possible to create custom event types and implement detection for it in the core code.
client.Subscribe(EventTypes.Events.FailedJobPeak, toEmails);
client.Subscribe(EventTypes.Events.ServerTimeout, toEmails2);

List<INotifier> clientList = new List<INotifier>();
clientList.Add(client);

GlobalConfiguration.Configuration.AddNotificationSystem(clientList);

app.UseHangfireServer(new BackgroundJobServerOptions()
{
      JobCheckInterval = new TimeSpan(0, 5, 0),
      JobCheckThreshold = 20,
      ServerName = "Server One"
});

```

The notification store can then be accessed using the static method, NotificationStore.Current, in the same way as JobStorage.
This can then be used to call the NotifyAll method, which takes a EventType parameter and alerts all emails registered for that EventType in each registered client. For example,

```
//Server timeout detected in server watchdog
NotificationStore.Current.NotifyAll(EventType.Event.ServerTimeout. "Subject", "Message");
```

### DIFF
--- a/src/Hangfire.Core/BackgroundJobServer.cs
+++ b/src/Hangfire.Core/BackgroundJobServer.cs
@@ -195,7 +195,7 @@ namespace Hangfire
                 var name = _options.ServerName;
                 var msg = $"The server with machine name: \"{ name }\" has had a peak of failed jobs.";
 
-                NotificationStore.Current.NotifyAll(EventTypes.Events.FailedJobPeak, "Failed job peak", msg);
+                NotificationStore.Current.NotifyAll(EventTypes.FailedJobPeak, "Failed job peak", msg);
             }
         }
     }
--- a/src/Hangfire.Core/BackgroundJobServer.cs
+++ b/src/Hangfire.Core/BackgroundJobServer.cs
@@ -17,6 +17,7 @@
 using System;
 using System.Collections.Generic;
 using System.Linq;
+using System.Threading;
 using Hangfire.Annotations;
 using Hangfire.Client;
 using Hangfire.Common;
@@ -33,10 +34,8 @@ namespace Hangfire
 
         private readonly BackgroundJobServerOptions _options;
         private readonly BackgroundProcessingServer _processingServer;
-        private readonly object _failedLock;
 
         private DateTime _lastEmailNotification;
-        private bool _hasFirstFailedDateTime;
         private long _failedJobsCount;
 
         /// <summary>
@@ -89,10 +88,8 @@ namespace Hangfire
             if (additionalProcesses == null) throw new ArgumentNullException(nameof(additionalProcesses));
 
             _options = options;
-            _lastEmailNotification = new DateTime();
-            _hasFirstFailedDateTime = false;
+            _lastEmailNotification = DateTime.MinValue;
             _failedJobsCount = 0;
-            _failedLock = new object();
 
             var processes = new List<IBackgroundProcess>();
             processes.AddRange(GetRequiredProcesses());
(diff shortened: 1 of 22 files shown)


src/Hangfire.Core/BackgroundJobServer.cs
```diff
@@ -168,5 +180,32 @@ public void Start()
         public void Stop()
         {
         }
+
+        public void Update()
+        {
+            var threshold = _options.JobCheckThreshold;
+            var interval = _options.JobCheckInterval;
+
+            lock (_failedLock)
```
nirinchev: You could use `Interlocked.Increment` instead of the manual lock.

src/Hangfire.Core/BackgroundJobServer.cs
```diff
@@ -168,5 +180,32 @@ public void Start()
...
+            {
+                _failedJobsCount++;
+            }
+
+            if (!_hasFirstFailedDateTime || DateTime.Now.Subtract(interval) > _lastEmailNotification)
```
nirinchev: You don't need the `_hasFirstFailedDateTime` - simply initialize `_lastEmailNotification` with `DateTime.MinValue` so the second check will be always true the first time.

src/Hangfire.Core/IWorkerObserver.cs
```diff
@@ -0,0 +1,13 @@
...
+using System.Collections.Generic;
+using System.Linq;
+using System.Text;
+using System.Threading.Tasks;
+
+namespace Hangfire
+{
+    public interface IWorkerObserver
+    {
+        void Update();
```
nirinchev: This doesn't seem like a good name - you are calling `Update` only if the worker fails.

Anthon123: The other option is to call it Notify but this is already used in other files relevant to this functionality, it is called Update to cause less confusion. It updates the background job server to notify it about the event of a new failed job.

src/Hangfire.Core/Notification/EventTypes.cs
```diff
@@ -0,0 +1,17 @@
+﻿using System;
+using System.Collections.Generic;
+using System.Linq;
+using System.Text;
+using System.Threading.Tasks;
+
+namespace Hangfire.Notification
+{
+    public class EventTypes
```
burningice2866: Why have a class containing your enum? The normal thing would be to add the enum directly beneath the namespace as a self-contained type.

burningice2866:
I like the idea of "Notifications", but the Email Module part i would keep out of the equation, maybe release it as a package maintained by your own organization.

src/Hangfire.Core/IWorkerObserver.cs
burningice2866: I suppose you technically could listen to failed jobs using IApplyStateFilter and listen to when a job enters the Failed state,

The whole Notification system could be made that way as well, not changing any existing infrastructure in core, but listen to all State changes and fire off your notification-methods.

odinserj:
Monitoring of background processing should be performed using logging abstractions. For all modern logging framework it is possible to send log messages via SMTP. So this feature is very doubtful, especially since it's so complex and require a lot of additional abstractions.

PR #615, opened 2016-07-15
Requeue to specific queue through dashboard

stevetsang:
As a developer it can be difficult to debug a job through a live system.  We have added functionality to the dashboard to give you an option to requeue to a different queue for jobs that have failed (also works for jobs that have succeeded).  

We can then create a debug server, connect it to live, then requeue the failed job against the debug queue to debug the error.


pieceofsummer:
If multiple servers handle the same queue, then id attribute won't be unique, which produces incorrect HTML. Also, queue name may conflict with some other DOM element id. This may break any id lookups in JS.

PR #617, opened 2016-07-19
Fix Cron.cs for recurring jobs using intervals

stevetsang:
When scheduling Cron Jobs using the Interval functions in the Cron class it seems to just schedule the job for a minute + the interval time.  Due to this the schedule job seems to execute just every minute e.g. If we call the HourInterval with 12 it would show up on the dashboard as 1minute, 12hours and it would just execute every minute.


odinserj:
Thanks guys for catching this! Could you please target it against the `master` branch? `master` branch is for patch releases, `dev` branch is for next minor releases with new features.

stevetsang:
ok will close and retarget against master!

PR #618, opened 2016-07-19
Fix Cron.cs for recurring jobs using intervals

stevetsang:
When scheduling Cron Jobs using the Interval functions in the Cron class it seems to just schedule the job for a minute + the interval time. Due to this the schedule job seems to execute just every minute e.g. If we call the HourInterval with 12 it would show up on the dashboard as 1minute, 12hours and it would just execute every minute.


odinserj:
Thank you, @mahoneyj2 and @stevetsang! Merging, will ship it soon :ship:

Issue #611, opened 2016-07-14
Task executes is put in queue more than once when triggering manually


evaldas-raisutis:
When you trigger a task manually from the UI, if you click the button "Trigger Now" multiple times quickly, selected tasks will be sent to queue multiple times before deselecting tasks and disabling the button. I would expect the tasks to be sent once, regardless how many times the button is clicked.

odinserj:
PR is welcome! :smile:

odinserj:
Fixed by PR in 1.6.1

PR #626, opened 2016-07-26
Fixed UI command behaviour

mahoneyj2:
Fixed UI command behaviour across all pages, such that buttons can only be pressed once per command. This removes side effects such as being able to trigger the same job multiple times on the recurring jobs page. #611

### DIFF
--- a/src/Hangfire.Core/Dashboard/Content/js/hangfire.js
+++ b/src/Hangfire.Core/Dashboard/Content/js/hangfire.js
@@ -457,9 +457,9 @@
 
                         $.post($this.data('url'), { 'jobs[]': jobs }, function () {
                             clearTimeout(loadingDelay);
-                            $this.button('reset');
                             window.location.reload();
                         });
+                        $this.button('reset');
                     }
 
                     e.preventDefault();
--- a/src/Hangfire.Core/Dashboard/Content/js/hangfire.js
+++ b/src/Hangfire.Core/Dashboard/Content/js/hangfire.js
@@ -335,13 +335,13 @@
                 var confirmText = $this.data('confirm');
 
                 if (!confirmText || confirm(confirmText)) {
+                    $this.prop('disabled');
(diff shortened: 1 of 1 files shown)


odinserj:
Hi @mahoneyj2, how this fixes the problem? The linked issue is more related to a delay between the click and `loading` state change (due to `setTimeout` call). It is better to make buttons disabled before making a call to `setTimeout`. Btw, original issue says that other buttons (such as "Trigger now") are affected.

mahoneyj2:
Sorry for the confusion @odinserj! 

The problem which I could identify from the issue was that when you select multiple jobs and "Trigger now", the button is disabled for a short period during which the command is dealt with, however the button is re-enabled before the page is reloaded after the POST. This meant that you could press the button multiple times, with the same jobs selected, causing the same command to be issued to the back-end. In the case of the recurring jobs page, this meant the "Trigger now" could be pressed before the page was reloaded, causing the same job to be triggered multiple times. 

This behaviour is the same across all the pages, although only recurring jobs causes a problem, as you cannot re-queue jobs which are already queued.

I'll have another look, but it seemed to solve all the problems on my end!

odinserj:
Ah, now I see. What do you think about removing the `button('reset')` method call at all and disabling the button before calling the `setTimeout` method to prevent from clicking a button before receiving a POST response?

mahoneyj2:
Yeah that sounds like a better idea, just realised it was pointless as the page will be reloaded anyway. The earlier the button is disabled the better! Will update to that now!

odinserj:
Nice! Would you like to update the other button as well? It is also in the `hangfire.js` file, with the same code.

mahoneyj2:
Didn't spot that before! Have updated it now. Sorry for being a little slow, one of the first times using GitHub properly!

odinserj:
Nevermind, @mahoneyj2, sometimes I'm much slower :smile: Thank you for the contribution!

PR #659, opened 2016-09-02
Add Every X minute method to Cron.

ojosdegris:


odinserj:
@ojosdegris, this change is controversial, because method's semantics is inconsistent with the other `*ly` methods – none of them are using the *every X* semantics. This change also requires the other methods (every N hours, days, etc.) to be added as well, but I don't want to make the `Cron` class huge.

ojosdegris:
I agree,  however that is thepattern we are using the most right now. thought it might be usefull for others.
we can rename it to XMinutely or add simple Every method accepting TimeSpan.
happy to do pull request if you think it is more apropriate.

odinserj:
Ohh, the method [`MinuteInterval`](https://github.com/HangfireIO/Hangfire/blob/129707d66fde24dc6379fb9d6b15fa0b8ca48605/src/Hangfire.Core/Cron.cs#L221) is already here :smile: The logic is the same.

odinserj:
You can add argument validation for all of these methods if you'd like – I will be happy to merge that PR.

ojosdegris:
Oh that is akward, that's exactly what I was looking for and I did not see all those methods. thanks.

ojosdegris:
I see what happens. Code i was working on is on older version of Hangfire (1.5) and it do not have thise https://github.com/HangfireIO/Hangfire/pull/618 PR.

odinserj:
So, closing this PR, as the feature was already implemented.

PR #660, opened 2016-09-06
Add entry point for custom job naming strategies

pieceofsummer:
Because of shortcomings of `DisplayNameAttribute` (or lack thereof on .NET Core), I'd like to propose another approach to job naming: `IDashboardJobNameProvider`.

Like `IDashboardAuthorizationFilter`s, name provider is configured in `DashboardOptions` during startup, and implements a single method taking current `DashboardContext` (so you have access to Owin environment or .NET Core's `HttpContext`) and the `Job` of interest. 

Here's an example of what it looks like:

``` c#
public class SampleJobNameProvider : IDashboardJobNameProvider
{

    public string GetName([NotNull] DashboardContext context, [NotNull] Job job)
    {
        // You may implement different approaches for job naming, be it an attribute ...

        string jobName = job.Method.GetCustomAttribute<DisplayNameAttribute>()?.DisplayName;
        if (jobName == null)
        {
            // ... or something else you might think of ...

            jobName = Regex.Replace(job.Type.Name, "([a-z])([A-Z])", "$1 $2");
        }

        // ... then you might want to translate it into user's preferred UI language ...

        var localizerFactory = context.GetHttpContext().RequestServices
                                      .GetService<IStringLocalizerFactory>();
        if (localizerFactory != null)
        {
            var baseName = job.Type.FullName;
            var location = job.Type.GetTypeInfo().Assembly.GetName().Name;

            jobName = localizerFactory.Create(baseName, location)[jobName];
        }

        // ... or maybe apply some formatting afterwards ...

        try
        {
            return string.Format(jobName, job.Args.ToArray());
        }
        catch
        {
            return jobName;
        }
    }

}
```

### DIFF
--- a/src/Hangfire.Core/Dashboard/HtmlHelper.cs
+++ b/src/Hangfire.Core/Dashboard/HtmlHelper.cs
@@ -106,6 +106,16 @@ namespace Hangfire.Dashboard
                 return Strings.Common_CannotFindTargetMethod;
             }
 
+            var jobNameProvider = _page.JobNameProvider;
+            if (jobNameProvider != null)
+            {
+                var jobName = jobNameProvider.GetName(_page.Context, job);
+                if (!string.IsNullOrEmpty(jobName))
+                {
+                    return jobName;
+                }
+            }
+
 #if NETFULL
             var displayNameAttribute = job.Method.GetCustomAttribute(typeof(DisplayNameAttribute)) as DisplayNameAttribute;
             if (displayNameAttribute != null && displayNameAttribute.DisplayName != null)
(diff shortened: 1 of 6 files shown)


rsilva93:
@odinserj Any idea of if/when is this going to production? It's a very nice feature for net core developers, cause today we're sad with that method namespaces showing on dashboard :(

odinserj:
Alex, I've just merged this into the `dev` branch, and will release 1.7.0-beta1 today with these changes :heart:

pieceofsummer:
@odinserj great! Would definitely give it a try!

odinserj:
I'll mark it with the 1.7.0 milestone then and check a bit later ("later" sounds dangerous, he-he) 😃

PR #674, opened 2016-09-23
fixed bug 666 with exception on showing detailes of  Continuation wit…

MaksimSimkin:
…h JsonSerializerSettings   TypeNameHandling = TypeNameHandling.Objects

It's a fast and working fix. I am not  100%  happy with making  a private class to public, but ibn my opinion it's better than havin two identical private classes, for derialization and for deserialization


odinserj:
@MaksimSimkin, awesome, thanks for contributing! I just thought to start working on this problem 😃

PR #684, opened 2016-10-03
Fixes #595

### LINKED ISSUE #595: Recurring job '?' can not be scheduled due to job load exception.
Talking about Scalability/Distrubted Systems I think each server might see only its queues.
In my production envrionment I'm running two instances in two servers. Each server actually runs over different queues as below
image

Server "A" has only libraries needed there and in the other server ("B") only the libraries needed also. When server "B" tries to Schedule a job it throws `JobLoadException` because there's no library into its binary folder. What I suggest is to change few classes to restrict `RecurringJobScheduler` class to watch only the right queues.

The queue is already in the context and could be used to filter recurring-jobs as well.
image

RupertAvery:


RupertAvery:
Any news on this? Is my fix valid?  Right now we're getting a problem that another instance of HangFire with a different queue is picking up Scheduled jobs meant for another instance. This tends to block the job from being processed on time, since it retries for a while before giving up.

RupertAvery:
Closing this PR as https://github.com/HangfireIO/Hangfire/pull/755 is better implemented

PR #686, opened 2016-10-07
Fixes #657

### LINKED ISSUE #657: Add yet another validation for the timeout parameter in SqlServerDistributedLock
The `timeout` parameter is already validated to fit the `SqlCommand.CommandTimeout` property, but it is not validated against the `sp_getapplock`'s `@LockTimeout` parameter. The latter is in milliseconds, and should be less or equal to `Int32.MaxValue`.

So, to prevent the confusion (like in #647), we can add the `timeout.TotalMilliseconds < Int32.MaxValue` check to the `SqlServerDistributedLock`'s constructor and throw exception with meaningful message, when check fails.

**Optional:** Add validation for the `DisableConcurrentExecution`'s `timeoutInSeconds` parameter as well.

t0mburton:
Throw argument exception when timeout milliseconds exceeds int.max. Prevents issue with sp_getapplock's @LockTimeout param.

### DIFF
--- a/src/Hangfire.SqlServer/SqlServerDistributedLock.cs
+++ b/src/Hangfire.SqlServer/SqlServerDistributedLock.cs
@@ -1,5 +1,5 @@
 // This file is part of Hangfire.
-// Copyright © 2013-2014 Sergey Odinokov.
+// Copyright Â© 2013-2014 Sergey Odinokov.
 // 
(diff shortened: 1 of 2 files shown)


odinserj:
After the long delay, it was finally merged. Thank you @t0mburton for the contribution! :+1: And sorry for the delay 😟

PR #687, opened 2016-10-07
Customize CSS for Dashboard UI

plmwong:
Refer to https://discuss.hangfire.io/t/customize-css-for-ui/1141

As described in the above discussion, added `ExtraCssUrl` parameter to `DashboardOptions`, which will be added into the `LayoutPage` as a linked stylesheet if it has been set to a non-null value.


pieceofsummer:
Providing extra CSS is a great feature, but it would be a pity if it can't be used by extensions. Would you like to fix merge conflicts and modify the PR to allow multiple CSS (and also JS) files?

Also, the PR contains a lot of meaningless changes (like removing trailing spaces etc.) which should be excluded from the commit.

PR #709, opened 2016-10-28
Fix BeginScope() returning SimpleJobActivatorScope

### LINKED ISSUE #706: AspNetCoreJobActivator.BeginScope() returns SimpleJobActivatorScope
Since `BeginScope()` without args is not overridden in `AspNetCoreJobActivator`, the base implementation from `JobActivator` is being called. 

Even though it is obsolete, it'd be better to return a correct scope :)

pieceofsummer:
Fixes #706.


odinserj:
🚢ping!

PR #710, opened 2016-10-29
Make context constructors public

pieceofsummer:


odinserj:
Merging, thanks! Looks like all the Travis CI issues were finally eliminated. I've also tuned up @codecov-io comments to be more modest 😄

PR #711, opened 2016-11-01
Window resize cause errors and uncontrolled growth of event and poll listeners

Yarmonov:
## Preamble

In the ongoing project I was integrating Hangfire dashboard into the existing UI using simple iframe approach. To auto-size the iframe I used custom resizing handler and noticed that Hangfire throws a lot of exceptions.

## Steps to reproduce

1. Execute MvcSample and navigate to /hangfire URL (Chrome, Firefox, IE - all are affected).
2. Open browser's developer console.
3. Resize the window.
4. Move mouse over graphs on the dashboard.
5. Each mouse movement event will throw "Uncaught TypeError: Cannot read property 'getBoundingClientRect' of null(…)".
6. The same exception will be also thrown periodically, event without mouse moving.

## Problem

In the window.onresize handler (hangfire.js:221) existing graphs placeholders are emptied and new graphs created. During the creation phase, Rickshaw.Graph.HoverDetail adds listeners for mousemove and mouseout events of the graph placeholder. But it is just emptied, so on resize these listeners aren't removed and still exists. Therefor each resize adds 2 new listeners. Besides this listeners growth the old ones are working with detached nodes those and this causes an error mentioned above.

Another problem is caused by statistics poller. When realtim graph is created, it adds a listener. But when recreated during resize old listener is not removed, but new one is added.

## Solution

1. I have added graph placeholder node recreation, This removes obsolete handlers from it,
2. Instead of calling _createGraphs, I'm calling new Hangfire.Page creation. It calls _createGraphs and also recreate poller. So this removes obsolete poller listeners.
3. Slight fix in _createRealtimeGraph. It has a condition if placeholder exists, but not all code that manipulate with the placeholder is inside of this condition (just compare to _createHistoryGraph that has a proper wrap).

### DIFF
--- a/src/Hangfire.Core/Dashboard/Content/js/hangfire.js
+++ b/src/Hangfire.Core/Dashboard/Content/js/hangfire.js
@@ -89,7 +89,17 @@
         };
 
         RealtimeGraph.prototype.update = function() {
-            this._graph.update();
+            var graph = this._graph;
+
+            var width = $(graph.element).innerWidth();
+            if (width !== graph.width) {
+                graph.configure({
+                    width: width,
+                    height: 200
+                });
+            }
+
+            graph.update();
         };
 
         return RealtimeGraph;
@@ -134,7 +144,17 @@
         }
 
         HistoryGraph.prototype.update = function() {
-            this._graph.update();
+            var graph = this._graph;
+
+            var width = $(graph.element).innerWidth();
+            if (width !== graph.width) {
+                graph.configure({
+                    width: width,
+                    height: 200
+                });
+            }
+
+            graph.update();
         };
 
         return HistoryGraph;
@@ -204,9 +224,9 @@
         }
 
         Page.prototype._createGraphs = function() {
-            this.realtimeGraph = this._createRealtimeGraph('realtimeGraph');
-            this.historyGraph = this._createHistoryGraph('historyGraph');
-            
+            var realtime = this.realtimeGraph = this._createRealtimeGraph('realtimeGraph');
+            var history = this.historyGraph = this._createHistoryGraph('historyGraph');
+
             var debounce = function (fn, timeout) {
                 var timeoutId = -1;
                 return function() {
@@ -217,12 +237,9 @@
                 };
             };
 
-            var self = this;
             window.onresize = debounce(function () {
-                $('#realtimeGraph').html('');
-                $('#historyGraph').html('');
-
-                self._createGraphs();
+                realtime.update();
+                history.update();
             }, 125);
         };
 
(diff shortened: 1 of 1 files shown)


odinserj:
Nice catch, @Yarmonov, we know now who's producing these errors. But I still get these issues after fetching [this build](https://ci.appveyor.com/project/odinserj/hangfire-525/build/1.6.7-build-02318-pyjoddsg):

```xml
<package id="Hangfire.Core" version="1.6.7-build-02318" targetFramework="net452" />
<package id="Hangfire.MemoryStorage" version="1.3.0.0" targetFramework="net452" />
```

image

Yarmonov:
@odinserj, it's strange. I get the build, installed its packages into empty project and things looks fine. Can you check, please, that cache is emptied and/or ensure that there is a "_deleteGraph" function for instance in the "/hangfire/js167" script.

If JS code is actual, can you provide me steps to reproduce (maybe there are some details those differs from the mine) and environment (browser)?

Yarmonov:
@odinserj It seems I've found an additional circumstances those can lead to these issues. Have added statistics poller stopping before graphs deletion.

odinserj:
Thanks for highlighting the problem! `HoverDetail` mouse listeners doesn't add any method to remove the registered event listeners. There is even a [PR for this](https://github.com/shutterstock/rickshaw/pull/477), but it wasn't merged since 2014.

The root of the problem is improper actions when a window resize occurred. Rickshaw does support resizes without recreating anything, please see the [resize.html](https://github.com/shutterstock/rickshaw/blob/master/examples/resize.html) demo. Have you tried to use this example?

Yarmonov:
Exactly. But the same problem is for hangfire.StatisticsPoller class as well. It also doesn't remove listeners. As a result we get two kinds of the same error. The first one caused by Rickshaw happens on mouse move and the second one, caused by Hangfire happens regularly (depends on poller's delay).
		
Yes, I see the resize.html demo, but circumstances differs. It works with the same graph, meanwhile Hangfire is recreating graphs. So events attached to the old instances are generating errors.

odinserj:
We can simply change the `window.onresize` function to not to recreate graphs, replacing its contents by example of [resize.html](https://github.com/shutterstock/rickshaw/blob/master/examples/resize.html). When we remove the recreation logic, there will be no new listeners, and no obsolete ones. It's recreation logic causes these errors, and if we change recreation with resizing, all the errors will go away (debounce function may also be removed).

I've originally did graph recreation by an accident, seems like I didn't find `resize.html` demo at that moment.

```is
            window.onresize = debounce(function () {
                $('#realtimeGraph').html('');
                $('#historyGraph').html('');

                self._createGraphs();
            }, 125);
```

Yarmonov:
You're right. I've completely redesigned the solution and also applied some minor fixes.

odinserj:
Awesome, Stanislav, I don't see errors anymore. Finally this bug was fixed after 2 years of existence, thanks for your hard work :+1: 💪 !

PR #712, opened 2016-11-02
Add tests covering GetNonOpenMatchingMethod method of TypeExtensions class

aidmsu:
I added some tests covering GetNonOpenMatchingMethod method of TypeExtensions class.


tests/Hangfire.Core.Tests/Common/TypeExtensionsFacts.cs
```diff
@@ -42,22 +45,162 @@ public void ToGenericTypeString_PrintsClosedGenericNestedClassName_WithGivenType
...
+        public class NestedNonGenericClass
+        {
+            public class DoubleNestedGenericClass<T1, T2, T3>
+            {
+
+            }
+        }
+    }
+
+    public interface IChild
```
odinserj: Consider switching `IParent` / `IChild` interfaces, because parent interface can't implement a child one.

tests/Hangfire.Core.Tests/Common/TypeExtensionsFacts.cs
```diff
@@ -42,22 +45,162 @@ public void ToGenericTypeString_PrintsClosedGenericNestedClassName_WithGivenType
...
+        }
+
+        [Fact]
+        public void GetNonOpenMatchingMethod_ReturnsCorrectGenericMethod()
+        {
+            var method = TypeExtensions.GetNonOpenMatchingMethod(typeof(NonGenericClass), "GenericMethod",
+                new Type[] { typeof(int), typeof(string), typeof(object) });
+
+            Assert.Equal("GenericMethod", method.Name);
+            Assert.Equal(typeof(NonGenericClass), method.DeclaringType);
```
odinserj: Please add an assertion that proves that resulting method is a generic one, but doesn't contain any generic parameters (i.e. it is a closed generic method).

src/Hangfire.Core/Common/TypeExtensions.cs
```diff
@@ -12,14 +12,15 @@
...
-using System.Text.RegularExpressions;
-
+// License along with Hangfire. If not, see <http://www.gnu.org/licenses/>.
+
+using System;
+using System.Collections.Generic;
+using System.Linq;
+using System.Reflection;
+using System.Text.RegularExpressions;
+using Hangfire.Annotations;
```
odinserj: Why are these lines are shown on the diff? Looks like something is wrong with line endings, either in the old code, or in the new one. If old code is wrong, it's better to keep these changes.

PR #713, opened 2016-11-03
Fix bug related to incorrect choosing overloading method

aidmsu:
Fixed a bug related to the method `GetNonOpenMatchingMethod` returns incorrect method from existing overloaded methods. So if you have two methods:
```csharp
void Method(object arg) { }
void Method(SomeClass arg) {}
```
And you try enqueue job:
```csharp
var someObject = new SomeClass();
BackgroundJob.Enqueue(() => Method(someObject);
```
You expect invocation of `void Method(SomeClass arg)` method but Hangfire invokes `void Method(object arg)`.

### DIFF
--- a/src/Hangfire.Core/Common/TypeExtensions.cs
+++ b/src/Hangfire.Core/Common/TypeExtensions.cs
@@ -169,7 +169,7 @@ namespace Hangfire.Common
                     continue;
                 }
 
-                return method2ParameterType.IsAssignableFrom(method1ParameterType)
+                return method2ParameterType.GetTypeInfo().IsAssignableFrom(method1ParameterType.GetTypeInfo())
                     ? overloadedMethod1
                     : overloadedMethod2;
             }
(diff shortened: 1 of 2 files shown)


odinserj:
For non-generic methods, the result method should be chosen **only** on the basis of `parameterTypes` argument. So strict equality should be used here instead of calling to `IsAssignableFrom` method. Moreover, if there are no candidates that strictly satisfy the types given in `parameterTypes` array, `null` result should be returned:

```csharp
class Test
{
    public void Method(object value) {}
}

var result = typeof(Test).GetNonOpenMatchingMethod("Method", new [] { typeof(Test) });
Assert.Null(result);
```

The `parameterTypes` array values is generated using the `MethodInfo.GetParameters` method, and not on the basis of argument types that were specified during the background job creation. So yes, there is a bug, but it relates to `IsAssignableFrom` method call instead of `Equals`, and not to wrong selection if there are overloads – `parameterTypes` are all about what overload should be exactly chosen.

odinserj:
Awesome, @aidmsu, will merge this after CI checks 👍

PR #648, opened 2016-08-18
Add Delay property to AutomaticRetryAttribute

FFLSH:

PR #714, opened 2016-11-03
adding possibility to set retry delay for failed jobs.

pavlo-p:
I've added a possibility to set RetryDelayInSeconds for AutoRetryAttribute.


FFLSH:
Hi Pavlo,

Looks like we've both added the same functionality, as I have a PR out (#648) which also allows a user to add a delay.

PR #719, opened 2016-11-07
Add support for complex generic parameters

### LINKED ISSUE #408: Generic method in Job
I'm trying to enqueue a generic method:

`BackgroundJob.Enqueue(() => MyService.DoWork<T>(p1, p2, p3);`

This seems to be a resolved issue: https://github.com/HangfireIO/Hangfire/issues/136 but I get the following error:

<pre>System.ArgumentException

The type or method has 1 generic parameter(s), but 0 generic argument(s) were provided. A generic argument must be provided for each generic parameter. 

System.ArgumentException: The type or method has 1 generic parameter(s), but 0 generic argument(s) were provided. A generic argument must be provided for each generic parameter.
   at System.RuntimeType.SanityCheckGenericArguments(RuntimeType[] genericArguments, RuntimeType[] genericParamters)
   at System.Reflection.RuntimeMethodInfo.MakeGenericMethod(Type[] methodInstantiation)
   at Hangfire.Storage.InvocationData.GetNonOpenMatchingMethod(Type type, String name, Type[] parameterTypes)
   at Hangfire.Storage.InvocationData.Deserialize()</pre>

What am I missing here?

Thanks

aidmsu:
Add support for generic parameters like List<T> to method GetNonOpenMatchingMethod of TypeExtensions class
Fixes  #408 
Now, Hangfire can enqueue generic methods. So you can write code:
```csharp
public void Foo<T>(IEnumerable<T> arg) ;

...

IEnumerable<int> aList = ...;
BackgroundJob.Enqueue(() => Foo(aList);
```
Also you can enqueue "complex" generic methods:
```csharp
public void Foo<T>(IEnumerable<IList<T>> argList, T arg) ;

...

IEnumerable<IList<int>> aList = ...;
int arg = ...;
BackgroundJob.Enqueue(() => Foo(aList, arg);
```


tests/Hangfire.Core.Tests/Common/TypeExtensionsFacts.cs
```diff
@@ -163,17 +163,136 @@ public void GetNonOpenMatchingMethod_HandlesMethodNameIsCaseSensitive()
...
+
+        [Fact]
+        public void GetNonOpenMatchingMethod_HandlesMethodHasParameterWhoseTypeContainsGenericParameterAndIsComplicated()
+        {
+            var method = TypeExtensions.GetNonOpenMatchingMethod(typeof(NonGenericClass), "OtherGenericMethod",
+                new[] { typeof(List<IEnumerable<int>>) });
+
+            Assert.Equal("OtherGenericMethod", method.Name);
+            Assert.Equal(typeof(NonGenericClass), method.DeclaringType);
+            Assert.Equal(true, method.IsGenericMethod);
```
odinserj: The following asserts aren't required. Other test cases may also be simplified.

```csharp
Assert.Equal(typeof(NonGenericClass), method.DeclaringType); // OtherGenericMethod exists only in this type
Assert.Equal(true, method.IsGenericMethod); // It's a generic method
Assert.Equal(false, method.ContainsGenericParameters); // You are validating all the parameters explicitly
```

src/Hangfire.Core/Common/TypeExtensions.cs
```diff
@@ -98,13 +110,97 @@ public static string ToGenericTypeString(this Type type)
...
+            this Dictionary<Type, Type> genericArguments,
+            Dictionary<Type, Type> addingGenericArguments)
+        {
+            foreach (var genericArgument in addingGenericArguments)
+            {
+                if (genericArguments.ContainsKey(genericArgument.Key) && 
+                    genericArguments[genericArgument.Key] != null &&
+                    genericArguments[genericArgument.Key] != genericArgument.Value)
+                {
+                    throw new InvalidOperationException("This generic parameter has been already identified as different type");
```
odinserj: You can simply replace an old value, because it's impossible to get here.

src/Hangfire.Core/Common/TypeExtensions.cs
```diff
@@ -98,13 +102,74 @@ public static string ToGenericTypeString(this Type type)
 
                 // Return first found method candidate with matching parameters.
                 return methodCandidate.ContainsGenericParameters
-                    ? methodCandidate.MakeGenericMethod(genericArguments.ToArray())
+                    ? methodCandidate.MakeGenericMethod(methodGenericArguments.Values.ToArray())
```
odinserj: From [MSDN](https://msdn.microsoft.com/en-us/library/ekcfxy3x(v=vs.110).aspx):

> **Dictionary<TKey, TValue>.Values Property**
> The order of the values in the Dictionary<TKey, TValue>.ValueCollection is **unspecified**, but it is the same order as the associated keys in the Dictionary<TKey, TValue>.KeyCollection returned by the Keys property.

Values of an array should be sorted first, based on `GenericParameterPosition` property, current solution may work today, but will not work tomorrow.

odinserj:
Hey-hey, it's finally merged! 🎉

augugiacoEnta:
@odinserj Hi dude,

I have a problem with a generic method when i try enqueue this. My generic method receive two type parameters.
When I try enqueue this method, I got an exception like this:

Example:

`BackgroundJob.Enqueue( () => Test<T1,T2>(param1);`

_The type or method has 2 generic parameter(s), but 0 generic argument(s) were provided. A generic argument must be provided for each generic parameter._

Can you tell me if this fix apply for my problem?

Thanks a lot!

odinserj:
@augugiacoEnta, what is the type of the `param1` variable?

augugiacoEnta:
@odinserj It's an interface that have T1,T2 types as parameters types.

Finally, I did a workaround with reflection and it's works!

odinserj:
Yeah, this PR will fix your problem!

PR #731, opened 2016-11-15
Rethrow JSON exception instead of returning a null value for argument

odinserj:
JSON exception during argument deserialization may cause `ReferenceConverter` to be used to perform the deserialization (type converters are here due to backwards compatibility). Since it couldn't properly handle serialized things, it will return null instead of re-throwing an exception, resulting in hidden error.


odinserj:
Thanks @aidmsu, I'm so happy to contribute to an OSS project! 😅

PR #736, opened 2016-11-18
Split json serializer settings into core settings and arguments settings

aidmsu:
Split json serializer settings into:
* **core settings** - json serializer settings which are responsible for serialization and desiarilation internal Hangfire data
* **arguments settings** - json serializer settings which affect serialization/deserialization of enqueuing method arguments.

Additional:
* add **DB** to abbreviations to avoid Resharper warnings
* add `CleanJsonSerializersSettingsAttribute` which clean `ArgumentsSerializerSettings` and `CoreSerializerSettings` after test


odinserj:
Why add breaking changes and cause everyone to update `SetSerializerSettings` to `SetArgumentsSerializerSettings`, and `JsonHelper.*Json` to `JsonHelper.*ArgumentsJson`? And why allow to change core serializer settings, since they are internal?

aidmsu:
@odinserj I made some modifications. So this PR:
1. Doesn't change `SetSerializerSettings`, `ToJson`, `FromJson` methods behavior  for compatability reasons.
2. Adds `SetArgumentsSerializerSettings` and `SetParametersSerializerSettings` methods to configure serializing of job arguments and job parameters.
3. Marks `SetSerializerSettings` as **Obsolete** because in my opinion users shouldn't have possibility to modify core serializer settings.
4. Adds `Serialize*` and `Deserialize*` arguments/parameters methods and makes them internal because Hangfire.Core is responsible for serializing job arguments and parameters. It's not necessary make this methods public.

aidmsu:
@odinserj You give me an idea that this problem should be solved another way. I close the PR before I'll implement that.

PR #737, opened 2016-11-21
Make Queue name accessible from RecurringJobDto

swordfish6975:
useful for:

```
using (var connection = JobStorage.Current.GetConnection())
            {
                var recurrings = StorageConnectionExtensions.GetRecurringJobs(connection);
              
                foreach (RecurringJobDto recurring in recurrings)
                {

                    if(recurring.Queue == "APPNAME")
                    {

                        if (recurring.Id != "ID_IN_APP_DATABASE_USING_ORM")
                        {
                            RecurringJob.RemoveIfExists(recurring.Id);
                        }

                    }
                 
                }

            }
```

### DIFF
--- a/src/Hangfire.Core/Storage/RecurringJobDto.cs
+++ b/src/Hangfire.Core/Storage/RecurringJobDto.cs
@@ -23,6 +23,7 @@ namespace Hangfire.Storage
     {
         public string Id { get; set; }
         public string Cron { get; set; }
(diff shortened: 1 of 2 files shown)


odinserj:
Thanks for this PR, @swordfish6975! Looks like this property was missed by an accident. Will be shipped with 1.6.7.

PR #722, opened 2016-11-09
Enable autoconversion of files line endings

aidmsu:
Change .gitattributes to enable autoconversion of line endings of .sln, .csproj, .vbproj, .fsproj, .dbproj etc. files. 

It allows to avoid mixing up of line endings in files.

A few links on the subject:
1. [GitHub recommendations about working with line endgins](https://help.github.com/articles/dealing-with-line-endings/).
2. [GitHub repo with .gitattributes files applicable to various programming languages and IDEs](https://github.com/alexkaratarakis/gitattributes)

PR #739, opened 2016-11-22
Finally fixes line endings problem. 

aidmsu:
This PR converts line endings in some files from CRLF to LF. I used [Charles Bailey's post on Stack Overflow](http://stackoverflow.com/questions/1510798/trying-to-fix-line-endings-with-git-filter-branch-but-having-no-luck/1511273#1511273).

There were two PRs partially solved line-ending problem:

* PR #721 fixed files which have both crlf and lf line endings. Line endings in these files were converted to CRLF.
* PR #722 enabled autoconversion of files line endings. So line endings in new files will be autoconverted to LF.

But there is a problem with files which have been committed with CRLF. Their line-endings are CRLF. Here is quote from git documentation
> Set to string value "auto"
> When text is set to "auto", the path is marked for automatic end-of-line conversion. If Git decides that the  content is text, its line endings are converted to LF on checkin. **When the file has been committed with  CRLF, no conversion is done**.

So if you want to modify file with CRLF line-endings (e.g. `RecurringJobManager`) you change some lines. But after it git will change line endings in this file to LF. And github will show that whole file was changed. 

This PR resolves problem described above.


odinserj:
Looks like all the problems related to EOL characters are gone! Great! 👍

Issue #734, opened 2016-11-17
Error 500 - View Dashboard Jobs


djfoz:
Hi,

Found a minor issue when viewing a job that can cause the dashboard to return an error 500 when viewing the status of a job. It occurs when a method accepts an array of reference types, where one of the values of the array is null.

![image](https://cloud.githubusercontent.com/assets/10595414/20409285/f5bdef34-ad10-11e6-8ad8-eb44dc8cd364.png)

JobMethodCallRenderer.cs line 133. item.ToString() results in a null reference exception. item?.ToString() appears to fix it. Although, argumentRenderer.Render looks to already to check. So looks like it could be checking for a null ref twice.

Thanks!

odinserj:
Nice catch, @djfoz! Would you like to create a PR with the proposed fix? Not a problem if it's checked twice, because we can't call `ToString` on `null` objects.

djfoz:
@odinserj - Sure will do!

Thanks

odinserj:
Looks like this bug is the latest for the [1.6.7 milestone](https://github.com/HangfireIO/Hangfire/milestone/52) :tada:! We'd like to make the 1.6.7 release tomorrow, could you tell me do you have time to fix this till tomorrow?

djfoz:
@odinserj I will submit the PR within the hour.

odinserj:
Merged, thanks again!

PR #741, opened 2016-11-24
Fixes error 500 in the dashboard

djfoz:
Fixes error 500 in the dashboard when rendering an array of reference types where an element could be null.

#734


odinserj:
@djfoz, thanks a lot for speed 👍! Congratulations with your first contribution on GitHub! 🎉 🎈 🎆

PR #750, opened 2016-12-01
Split serializer settings into core and arguments

aidmsu:
This PR splits `SerializerSettings` settings into `CoreSerializerSettings` and `ArgumentsAndParametersSerializerSettings`.

If user set custom json serializer setting via `SetSerializerSettings` method it won't affect serialization of core data: continuations, invocation data etc. This will affect only erialization/deserialization of job arguments and job parameters.


odinserj:
This PR is replaced with #1358.

PR #754, opened 2016-12-02
Typo - "Nexts jobs" should be "Next jobs"

danielabbatt:
Typo - "Nexts jobs" should be "Next jobs"


odinserj:
Thanks @danielabbatt, merging!

Issue #748, opened 2016-11-30
Support to specify the queue name


dbones:
when applying any operation except recurrent, we cannot specify the queue name.

Example:

```
BackgroundJob.Schedule(
    () => Console.WriteLine("Hello, world"),
    TimeSpan.FromDays(1));
```

Something like this would be great:

```
BackgroundJob.Schedule(
    () => Console.WriteLine("Hello, world"),
    TimeSpan.FromDays(1), 
    "application-1");
```

closest existing ticket - https://github.com/HangfireIO/Hangfire/issues/228

dbones:
this work-around looks to work (I have done the scheduled for now.)

**Client**
**1)** provide more data in the ScheduledState, put in a location (ie classlib/dll) which the client can access (does not matter much about the background server)

```
 public class ScheduledState : IState
    {
        public static readonly string StateName = "Scheduled";

        public ScheduledState(TimeSpan enqueueIn, string queue = "default")
            : this(DateTime.UtcNow.Add(enqueueIn), queue)
        {
        }

        [JsonConstructor]
        public ScheduledState(DateTime enqueueAt, string queue = "default")
        {
            EnqueueAt = enqueueAt;
            ScheduledAt = DateTime.UtcNow;
            Queue = queue;
        }

        public DateTime EnqueueAt { get; }
        public DateTime ScheduledAt { get; }
        public string Queue { get; }
        public string Name => StateName;
        public string Reason { get; set; }
        public bool IsFinal => false;
        public bool IgnoreJobLoadException => false;

        public Dictionary<string, string> SerializeData()
        {
            return new Dictionary<string, string>
            {
                { "Queue", Queue },
                { "EnqueueAt", JobHelper.SerializeDateTime(EnqueueAt) },
                { "ScheduledAt", JobHelper.SerializeDateTime(ScheduledAt) }
            };
        }  
    }
```

**Server**
**2)** add a filter (i chose to use an elect one)

```
public class UseQueueFromScheduledFilter : IJobFilter, IElectStateFilter
    {
        public void OnStateElection(ElectStateContext context)
        {
            
            var enqueuedState = context.CandidateState as EnqueuedState;
            
            if (enqueuedState != null )
            {
                var stateData = context.Connection.GetStateData(context.BackgroundJob.Id);
                string queueName;
                if (stateData.Data.TryGetValue("Queue", out queueName))
                {
                    enqueuedState.Queue = queueName;
                }
            }
        }
```

**server**
[...]

Issue #406, opened 2015-07-27
Requeueing a job without QueueAttribute forgets entered queue 


janpieterz:
When a job that doesn't have a QueueAttribute on its handler gets re queued, it is placed in the default queue instead of its originator queue. 

I use the following code to add a job:
```
foreach (var client in clients)
{
    BackgroundJobClient.Create<IManageClientUsers>(x => x.AddUser(user), new EnqueuedState(client.QueueId));
}

This is probably since it's a little bit sneaky to work around the specific queue problem, but I'd preferably not put attributes on every handler that I'd like to be routed (in my case I'd like everything to be routed to specific queues).

Possibly related to the solution of #228

odinserj:
That button places a job to the `default` queue, but `QueueAttribute` is able to change it. I think we should add a dropdown to the dashboard ui to choose what queue to use. Would you like to implement this feature (I can detail it)?

janpieterz:
Yeah, if you could guide me a little bit I'm more than happy to take it on.
Would you be open for saving the originator queue so that you don't necessarily need to select a dropdown? In case the 'farm' grows bigger and certain pieces fail, I can see the need to requeue a lot of messages not to a specific queue, but to the queues they came from/are supposed to go in.

More than willing to take either or both options on, if you could guide me which files/parts I should be digging in!

mhov:
@janpieterz you beat me to it.  I was having this same problem since I dynamically decide the queue names for jobs and ignore the "default" queue all together.  @odinserj , the Queue dropdown would be nice, but can we start off with honoring the original queue property so it makes it into the next release?

gandarez:
I fixed the problem coding a few.

Please refer to my branch:
https://github.com/gandarez/Hangfire/tree/requeue-to-initial-queue

Issue #228, opened 2014-10-13
Determine the queue at runtime


Atrejoe:
I would like to define the queue of the task at runtime.

Now there's the QueueAttribute, but it takes a compile-time argument for the queue name.
(I saw that the overload for creating a scheduled task previously had a queue argument, which was removed)

In my case the queues can be defined in a management application which pass throw the same method for scheduling. Something like:

    RecurringJob.AddOrUpdate(
        task.Id.ToString(),
        () => service.Run(task.Id)
        , task.TriggerExpression);

I have different HangFire clients which should only pick up some of the queues (they may be hardcoded).

See the image below: I have some tasks (executables) that are only allowed to be performed on particular clients. Other tasks (Web requests) can normally be performed on any client.

![image](https://cloud.githubusercontent.com/assets/585091/4614670/7c65dd2a-52e2-11e4-8802-fbb751f3679e.png)

The database on the left is our proprietary database, the one on the right is the database containing HangFire.

Frudy:
:+1: 
I also have a scenario similar to this where being able to select the queue at runtime is required.
Adding this possibility would be much appreciated, and enable us to use the otherwise wonderful Hangfire product :-)

odinserj:
Have you seen this solution – http://discuss.hangfire.io/t/one-queue-for-the-whole-farm-and-one-queue-by-server/490/4?u=odinserj?

Frudy:
Hi, thanks for the suggestion.
I tried it out, but it did not solve my problem.

In my scenario we have two independent hangfire servers on the same physical HW that I want to process jobs from two different queues. The key here is that these hangfire servers share DB but not code. So a job scheduled for queue "1" can not be picked up by hangfire server processing queue "2" since they do not share DLLs and thus can not run code for jobs assigned to the other queue. The code is simply not there.

Cheers
/Freddy

Atrejoe:
For me this might work, I actually do have the same codebase at each runner. When I find the time I will try this, thanks.

Frudy:
Hi

I think the solution to my problem could be like:

1) Configure server to process jobs on queue "queue1"
                config.UseServer(new BackgroundJobServerOptions
                {
                    Queues = new[] { "queue1" },
                });
2) Let client schedule recurrent task on "queue1"
            RecurringJob.AddOrUpdate<TaskService>("TaskName", (s) => s.Execute(), Cron.Daily(...), "queue1");
[...]

PR #755, opened 2016-12-02
Make Job Scheduling Aware of Queues the Server is Responsible For

EProd-Rhansen:
In working with this product in a distributed architecture, we have noticed our logs becomming inundated with job load exceptions from servers that are trying to schedule jobs when they do not have a reference to the assembly that contains the type that contains the job's logic.

This series of changes aims to isolate job scheduling such that servers only attempt to schedule or enqueue jobs that are assigned to a queue that the server is responsible for. Additionally, schedulers will now lock at the queue level which allows mutliple schedulers to run simultaneously so long as they are working from different queues.

None of these changes are breaking and can roll into any environment that is currently using the version that this change set was branched from.

### DIFF
--- a/Hangfire.sln.DotSettings
+++ b/Hangfire.sln.DotSettings
@@ -3,5 +3,8 @@
 	<s:Boolean x:Key="/Default/CodeInspection/CodeAnnotations/NamespacesWithAnnotations/=Hangfire_002ERedis_002EAnnotations/@EntryIndexedValue">True</s:Boolean>
 	<s:Boolean x:Key="/Default/CodeInspection/CodeAnnotations/NamespacesWithAnnotations/=Hangfire_002ESqlServer_002EAnnotations/@EntryIndexedValue">True</s:Boolean>
 	<s:Boolean x:Key="/Default/CodeStyle/CodeFormatting/CSharpFormat/SPACE_WITHIN_SINGLE_LINE_ARRAY_INITIALIZER_BRACES/@EntryValue">True</s:Boolean>
+	<s:Boolean x:Key="/Default/Environment/SettingsMigration/IsMigratorApplied/=JetBrains_002EReSharper_002EPsi_002ECSharp_002ECodeStyle_002ESettingsUpgrade_002EAddAccessorOwnerDeclarationBracesMigration/@EntryIndexedValue">True</s:Boolean>
+	<s:Boolean x:Key="/Default/Environment/SettingsMigration/IsMigratorApplied/=JetBrains_002EReSharper_002EPsi_002ECSharp_002ECodeStyle_002ESettingsUpgrade_002EMigrateBlankLinesAroundFieldToBlankLinesAroundProperty/@EntryIndexedValue">True</s:Boolean>
+	<s:Boolean x:Key="/Default/Environment/SettingsMigration/IsMigratorApplied/=JetBrains_002EReSharper_002EPsi_002ECSharp_002ECodeStyle_002ESettingsUpgrade_002EMigrateThisQualifierSettings/@EntryIndexedValue">True</s:Boolean>
 	<s:String x:Key="/Default/FilterSettingsManager/CoverageFilterXml/@EntryValue">&lt;data&gt;&lt;IncludeFilters /&gt;&lt;ExcludeFilters&gt;&lt;Filter ModuleMask="Hangfire.Core.Tests" ModuleVersionMask="*" ClassMask="*" FunctionMask="*" IsEnabled="True" /&gt;&lt;Filter ModuleMask="Hangfire.SqlServer.Tests" ModuleVersionMask="*" ClassMask="*" FunctionMask="*" IsEnabled="True" /&gt;&lt;/ExcludeFilters&gt;&lt;/data&gt;</s:String>
 	<s:String x:Key="/Default/FilterSettingsManager/AttributeFilterXml/@EntryValue">&lt;data /&gt;</s:String></wpf:ResourceDictionary>
--- a/src/Hangfire.Core/AutomaticRetryAttribute.cs
+++ b/src/Hangfire.Core/AutomaticRetryAttribute.cs
@@ -141,6 +141,11 @@ namespace Hangfire
             set { lock (_lockObject) { _logEvents = value; } }
         }
 
+        /// <summary>
+        /// Gets the name of the queue that the job should be placed into on retry attempts.
+        /// </summary>
+        public string RetryQueueName { get; }
+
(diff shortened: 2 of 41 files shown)


EProd-Rhansen:
See:

#228 
#406 
#748

EProd-Rhansen:
Is there a general feel for the willingness to move forward with accepting this PR?

burningice2866:
Unfortunately adding optional parameters to the public API is a breaking change so as i see it, all public methods/constructors now taking an optional `queueName` parameter needs to be overloads instead for this to be feasible at all.

The reason for this is that optional parameters is a compiler trick, which injects the default value into the callee, therefor you'll get a `MissingMethodException` in existing code that haven't been recompiled against the updated methods.

https://lostechies.com/jimmybogard/2010/05/18/caveats-of-c-4-0-optional-parameters/

EProd-Rhansen:
If you are referring to the changes to RecurringJob, then I did not add optional parameters for queue name. They already existed. The only thing I did was change the default value.

The one place that I can recall where I might have added queue name to the method signature where it did not already exist was to the creation of fire-and-forget jobs, but I am pretty sure that I implemented that with overloads rather than optional parameters.

If you can point out where I've done as you've said, then I'd be happy to look at correcting it. I generally shy away from default parameters whenever I can as I feel that they just end up bloating the method signature, on top of the fact that it is usually a pretty good indication that you're violating the single responsibility principle somewhere.

Beyond that, the problem that you've outlined (as I understand it) would only show its ugly face if someone simply dropped the updated assemblies into their production bin folder. In my mind. the only time that this could logically happen is if you're in an environment that does not treat updates to 3rd party references as a new build for your own applications - which is a terrible practice in its own right. In disciplined shops, you don't even get to production without a passing CI build - period - so this would never be a "breaking" update for them. 

For a change to be a breaking one, it must necessitate a code change on your end. I've never (personally) considered the need to simply recompile my own code against an updated assembly as a breaking change due to the fact that I would NEVER perform the update in any other way.      =)

burningice2866:
@EProd-Rhansen Even changing the default value is a breaking change, since the original default value is already compiled into to existing libraries

And you're missing the scenario of a solution with multiple libraries depending on HangFire but build againts different version numbers.

(version numbers are random and only there to prove my point)

Solution is upgading to Hangfire 2.9 which contains these changed default values
 - Lib1 is built against hangfire 2.3
 - Lib2 is built against hangfire 2.5
 - Lib3 is built against hangfire 2.4

In this case yo can't update Hangfire for the solution without having to update all the libraries, and you don't control the release cycle of them.

burningice2866:
@EProd-Rhansen i'll admit that i haven't combed through all the code changes, i merely noted a signature change here https://github.com/HangfireIO/Hangfire/pull/755/files#diff-762b9c9bc1acba9c001d8544882b1d4aR34 and a range of changed default parameter values, which raised my flag since i know from experience that it can be a minefield.

After look more carefully i suppose one can reason, that libraries not being recompiled wont fail, but one will experience unexpected behavior since the fixed queue-logic is not going into effect.

EProd-Rhansen:
Well, I was busy writing a well thought out rebuttal when you went and ruined it by coming to the conclusion all on your own! I guess that's what I get for thinking I know everything, lol.

It is splitting hairs, admittedly, but the worst that would happen in this case is that the framework exhibited the exact same behavior before and after the update was performed. The layman would just assume that their application was somehow auto-magically still using the old code until they recompiled it, whereas someone in-the-know would guess that someone cheated by just changing the default value of a public API.

For the record, I did make a (valiant) attempt at removing the optional parameters completely in favor of discrete method signatures in that class. However; that class is already overload hell. It would require some pretty dramatic restructuring of the API to remove them completely.

hangfireCoder:
We also are currently facing this same exact issue and would like to see this implemented in the release.

jcoutch:
I would love to see this get into a build sooner than later.  I'm getting 101,000+ "job load exception" messages between 3 servers in a 24-hour timespan.  :-(

WilkoSki:
Any movement on this? I agree with @jcoutch we are seeing too many job load exceptions and it just creates unnecessary noise.

Z1ni:
Waking this up, our server logs are filled with this exception noise.

nkwinder:
We want the server which added the recurrent job (or a direct job) to be the one that will execute the job!
If i understand correctly this pull request will be a solution for that. 
Is there any other way to achieve this?
Thats an important issue for us.

thank you

Tobias-H:
I registered account just to request you fix this issue. This is a blocker for us. We really wish to use Hangfire in more than one service per environment (multiple queues).

nkwinder:
Actually this seems to be working for us (we don't use retries for failed jobs):

```

// on app start
app.UseHangfireServer(new BackgroundJobServerOptions
{
	Queues = new[] { TaskQueueName.GetServerQueueName() }
});

// when adding recurring jobs
RecurringJob.AddOrUpdate<TaskRunner>(jobId, x => x.Run(taskId, true, null), cronExp, TimeZoneInfo.Local, TaskQueueName.GetServerQueueName());

// when executing jobs without schedule
IBackgroundJobClient hangFireClient = new BackgroundJobClient();
EnqueuedState myQueueState = new EnqueuedState(TaskQueueName.GetServerQueueName());
hangFireClient.Create<TaskRunner>(x => x.Run(id, false, null), myQueueState);

// helper for queue name
public static class TaskQueueName
{
	public static string GetServerQueueName()
	{
		return Environment.MachineName.ToLower().RemoveSpecialCharacters();
	}
}

```

Tobias-H:
@nkwinder Strange that you claim this works for you since it contradicts the reason for this pull request (server only loading jobs in its queues).
@nkwinder Also, your code assumes you want to use same queue for all Hangfire servers on same machine, which is often not the case, at least for us.

david-peden-q2:
We have also run into this issue. Many people here are claiming that it creates log noise. And that is true. But that is not the real problem. The real problem is that as a system's architecture gets more distributed, the chances of a successful re-run are reduced. In other words, as you scale a micro-service architecture up (each with their own dedicated queue against a single database), you inversely reduce the likelihood that a job will ever be picked up and successfully run. This is a serious design flaw. Let's move this discussion beyond the philosophical argument over the merit of optional parameters and, if necessary, considering versioning the codebase so that we can get a solution in place.

Thank you.

@burningice2866 @EProd-Rhansen @odinserj

Z1ni:
It seems that no progress has been made as there are some merge conflicts.
What's the "official" standpoint on this feature and integrating it to the codebase @odinserj?

Issue #762, opened 2016-12-14
Thread setup and tear-down


dgreenbean:
I am using Hangfire with some low-level libraries that have initialization and cleanup code per thread.  Hangfire does ensure that each worker runs in its own thread, but there is no way to hook code into the thread start/stop.

odinserj:
Could you tell me, what do you want to set?

ср, 14 дек. 2016 г. в 20:31, dgreenbean <notifications@github.com>:

dgreenbean:
In my case, I have a third-party library with some low-level code that requires an initialization for each thread.

odinserj:
Understood, but could you please provide an example? You can just paste here all the lines of code you are using.

dgreenbean:
This is essentially what I am doing:
`
new Thread(() =>
{
  using (new Lib())
  {
    while (true)
    {
      // Process job.
    }
  }
});`

dgreenbean:
Maybe I misunderstood the question.  This is an example of what I would hope to be able to set:
`backgroundJobServerOptions.WorkerSetup = lib.Init();
backgroundJobServerOptions.WorkerTearDown = lib.Terminate();`

Does that make sense?

odinserj:
Still not, what `lib.Init` and `lib.Terminate` methods are doing? Can you show me their source code? Why not to call this code in your method without touching Hangfire at all?

dgreenbean:
I'm not entirely sure what these methods do, since they are implemented by a third-party and are proprietary.  I have confirmed, however, that they need to be called within the same thread that does the processing later.  They also cannot be called multiple times for a single thread, meaning one initialization per thread is required.

I have worked around this by maintaining a ThreadLocal cache for now and ignoring the termination call.  Since this is a somewhat common pattern for multi-threaded support in many of the proprietary low-level libraries out there, I figured it would be useful to have a simple hook for this in Hangfire.

odinserj:
Thanks for letting me know about this use case. Since workaround is available, we don't need to introduce any changes to the Hangfire.Core itself.

PR #763, opened 2016-12-14
Issue #762: Adding setup/teardown methods to background processes, co…

dgreenbean:
This is a sample change for adding support for per-thread setup/teardown.


odinserj:
As discussed in the #762, this is a rare use case, and workaround is available, so it's not necessary to include these changes to the Hangfire.Core. But anyway, thanks for the PR!

PR #765, opened 2016-12-16
Fix Scheduled column title

aidmsu:
Use for Scheduled column title localized value from Strings.resx instead of string 'ScheduledJobsPage_Table_Scheduled'.


odinserj:
Thanks!

PR #768, opened 2016-12-19
Wrap exceptions throwing during job disposal in JobPerformanceException			407adc9

aidmsu:
Disposable jobs are automatic disposed after an invocation. Any exception which was threw during disposing should be wrap in JobPerformanceException.


odinserj:
Hi, `JobActivatorScope` is not a good place to catch exceptions and wrap them with the `JobPerformanceException`, because job activator does not perform background jobs. It's a job performer's responsibility to perform jobs and handle exceptions appeared in a user code. So I think it's better to move the logic to the background job performer.

PR #769, opened 2016-12-20
Modify identity columns type to bigint in all tables.

### LINKED ISSUE #749: state table exceed int32 max.
we recently exceeded int32 max in state table.
In new version convert int to bigint for id column.

[https://discuss.hangfire.io/t/hangfire-job-id/2387/4?u=hasanaydogar](https://discuss.hangfire.io/t/hangfire-job-id/2387/4?u=hasanaydogar)

aidmsu:
Identity columns in all tables have *int* type. Now there will be error if rows count exceed int.Max value.

This PR solves this problem by setting *bigint* type for identity columns in all tables.

Fix #749 

I also added in this PR sql migration:
* Make IX_HangFire_JobParameter_JobIdAndName index unique.

### DIFF
--- a/src/Hangfire.SqlServer.Msmq/MsmqJobQueueMonitoringApi.cs
+++ b/src/Hangfire.SqlServer.Msmq/MsmqJobQueueMonitoringApi.cs
@@ -41,9 +41,9 @@ namespace Hangfire.SqlServer.Msmq
             return _queues;
         }
 
-        public IEnumerable<int> GetEnqueuedJobIds(string queue, int @from, int perPage)
+        public IEnumerable<long> GetEnqueuedJobIds(string queue, int @from, int perPage)
         {
-            var result = new List<int>();
+            var result = new List<long>();
 
             using (var messageQueue = new MessageQueue(String.Format(_pathPattern, queue)))
             {
@@ -58,7 +58,7 @@ namespace Hangfire.SqlServer.Msmq
                         var message = enumerator.Current;
                         if (message == null) continue;
 
-                        result.Add(int.Parse(message.Label));
+                        result.Add(long.Parse(message.Label));
                     }
 
                     if (current >= end) break;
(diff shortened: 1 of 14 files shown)


aidmsu:
I found a few bugs in this PR. I'm working to solve them.

aidmsu:
I have plans to add one more sql migration to this PR:
* Add filter to indexes declared on ExpireAt column to cover only non-null values.

zzapal:
I think this PR should not change int into long, but rather into GUID to increase SQL performance when creating a lot of new jobs.

burningice2866:
@zzapal that would def. be a breaking change :/. Other Storage providers are free to use another id-type though - the [C1 Storage Provider](https://github.com/burningice2866/Hangfire.CompositeC1) uses GUIDs as you mention.

odinserj:
@zzapal, exactly! But I want to use sequential GUIDs, because random GUIDs are terrible for performance. But there's a problem, since sequential GUID generation is based on P/Invoke call to Windows API, and don't know yet what to do in .NET Core on Linux/MacOS – generate random GUIDs or use GUID combs. The problem with the latter that it can theoretically conflict with random GUIDs (Guid.NewGuid).

@burningice2866, I know how to implement this without introducing breaking changes to the consuming code. Identifiers of existing jobs can be converted to their integer representation in string.

burningice2866:
@odinserj You're right... we just have to remember to handle the case then where external code still calls `IMonitoringApi.JobDetails` passing a string-int - when that happens we would have to convert that to the corresponding guid, instead of parsing it directly.

burningice2866:
@odinserj What about delegating the guid-creation to the Sql Server using `NEWSEQUENTIALID`, that way you don't have any .Net platform specific requirements.

odinserj:
@burningice2866 delegation requires additional roundtrips, and by using GUIDs I wanted to avoid that behavior at least in 2.0. In 1.X integer identity will still be used, because GUIDs will not introduce anything significant to justify the changes.

Closing this PR, please see #898.

PR #777, opened 2016-12-27
Run tests inside Mono environment again.

aidmsu:
We can again test Hangfire inside Mono environment after [nuget issue 3085](https://github.com/NuGet/Home/issues/3085) issue was closed.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -4,7 +4,14 @@
 language: csharp
 sudo: required
 dist: trusty
-mono: latest
+
+matrix:
+  include:
+    - dotnet: 1.0.0-preview2-003121
+      mono: none
+      env: DOTNETCORE=1
+    - mono: latest
+      env: IS_MONO=true
     
 os:  
   - linux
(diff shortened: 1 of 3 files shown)


.travis.yml
```diff
@@ -35,27 +31,22 @@ branches:
...
 script:
-  - travis_retry dotnet restore --disable-parallel
-  - dotnet test tests/Hangfire.Core.Tests -verbose
- 
-# Tests on Mono disabled – waiting for https://github.com/NuGet/Home/issues/3085
-# - travis_retry mono .nuget/NuGet.exe install .nuget/packages.config -OutputDirectory packages
-# - travis_retry mono .nuget/NuGet.exe install .
-# - xbuild /p:Configuration=Release /verbosity:minimal Hangfire.Mono.sln
-# - mono ./packages/xunit.runner.console.2.2.0-beta2-build3300/tools/xunit.console.exe ./tests/Hangfire.Core.Tests/bin/Release/Hangfire.Core.Tests.dll
+  - if test "$NET_PLATFORM" == "mono";
```
odinserj: I think it's better to remove these variables and test existence of `mono` or `dotnet` commands, please see http://stackoverflow.com/questions/592620/check-if-a-program-exists-from-a-bash-script.

.travis.yml
```diff
@@ -35,27 +31,22 @@ branches:
 # Make sure build dependencies are installed.
 before_install:
```
odinserj: Maybe it's better to rename this step to the `install`? Because now we don't have the latter.

aidmsu:
@odinserj I removed extra instructions from .travis.yml. We will have to use last version of dotnet when Hangfire move to new csproj format.

odinserj:
How clean, can't wait for conversion to the new csproj format 😄

PR #787, opened 2017-01-10
Catch DistributedLockTimeoutException in some background processes.

aidmsu:
Catch `DistributedLockTimeoutException` exceptions in `RecurringJobScheduler`, `DelayedJobScheduler`, `ExpirationManager` background processes.

If `DistributedLockTimeoutException` is thrown during running of these processes it doesn't mean that work  wasn't done. It just mean another Hangfire server did the work. Thus we should catch this exception.

In case of a large set of Hangfire servers it will make easier to read logs because they won't contain many records about `DistributedLockTimeoutException`.


aidmsu:
I closed this PR because travis didn't build its.

PR #789, opened 2017-01-10
Catch DistributedLockTimeoutException in some background processes.

aidmsu:
Catch `DistributedLockTimeoutException` exceptions in `RecurringJobScheduler`, `DelayedJobScheduler`, `ExpirationManager` background processes.

If `DistributedLockTimeoutException` is thrown during running of these processes it doesn't mean that work  wasn't done. It just mean another Hangfire server did the work. Thus we should catch this exception.

In case of a large set of Hangfire servers it will make easier to read logs because they won't contain many records about `DistributedLockTimeoutException`.

### DIFF
--- a/src/Hangfire.Core/Server/DelayedJobScheduler.cs
+++ b/src/Hangfire.Core/Server/DelayedJobScheduler.cs
@@ -19,6 +19,7 @@ using Hangfire.Annotations;
 using Hangfire.Common;
 using Hangfire.Logging;
 using Hangfire.States;
+using Hangfire.Storage;
 
 namespace Hangfire.Server
 {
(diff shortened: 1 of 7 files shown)


src/Hangfire.Core/Server/DelayedJobScheduler.cs
```diff
@@ -144,39 +145,55 @@ public override string ToString()
         private bool EnqueueNextScheduledJob(BackgroundProcessContext context)
         {
             using (var connection = context.Storage.GetConnection())
-            using (connection.AcquireDistributedLock("locks:schedulepoller", DefaultLockTimeout))
```
odinserj: Why did you remove the `using` statement? Why not to wrap the whole block with the `try/catch` statement?

src/Hangfire.Core/Server/RecurringJobScheduler.cs
```diff
@@ -115,33 +115,50 @@ public void Execute(BackgroundProcessContext context)
             _throttler.Throttle(context.CancellationToken);
 
             using (var connection = context.Storage.GetConnection())
-            using (connection.AcquireDistributedLock("recurring-jobs:lock", LockTimeout))
```
odinserj: The same question about the `using` block.

src/Hangfire.SqlServer/ExpirationManager.cs
```diff
@@ -67,7 +68,18 @@ public void Execute(CancellationToken cancellationToken)
 
                 _storage.UseConnection(connection =>
                 {
-                    SqlServerDistributedLock.Acquire(connection, DistributedLockKey, DefaultLockTimeout);
```
odinserj: Why not to wrap the whole `UseConnection` method with the `try/catch` block?

src/Hangfire.Core/Server/DelayedJobScheduler.cs
aidmsu: @odinserj  I'd like to show that `AcquireDistributedLock` can throw `DistributedLockTimeoutException` but not whole block.

aidmsu:
@odinserj I added logging timeout exceptions with the debug level.

unicrus:
Guys, please merge it.

odinserj:
Merged into `master` via 1d6ad9ae50d4011863dd5eba905ac2ad365ab979, 7f521e97ed8246895ef7fba7292445b541d5f0fc, 7c141c5f5c972f4069c1ff1d0c167535b442f64b, 5aed50a41d70c1b7d46cad7c2c49797e1a32b1a8, fecf74a1ff5dd90bb356980b0b7966863ea6a2a8 commits.

Issue #799, opened 2017-01-23
Queue parameter is ignored when scheduling job with IBackgroundJobClient


shatl:
Hi,
same job goes to different HF queues when enqueued using static method and IBackgroundJobClient

```
  // .NET Core controllerm _jobs injected in constructor 
 
        [HttpGet("{message}", Name = "Echo")]
        [Route("echo")]
        public string Echo(string message)
        {
            // goes to default queue
            /* IBackgroundJobClient*/ _jobs.Enqueue<ITestJob>(x => x.DoIt(message));
            // goes to correct queue
            BackgroundJob.Enqueue<ITestJob>(x => x.DoIt(message));
            return message;
        }

    public interface ITestJob
    {
        [Queue(Settings.HangfireTransmitQueueName)]
        void DoIt(string msg);
    }
    class TestJob : ITestJob
    {
        public void DoIt(string msg)
        {
            
        }
    }
```

pieceofsummer:
Wow, looks like one of the bugs I've just fixed in #800 :)

The reason is, `IBackgroundJobClient` uses `IJobFilterProvider` service from IoC (which is `GlobalJobFilters.Filters`), while the static wrapper uses `JobFilterProviders.Providers`. The latter combines global job filters from the former one with filters coming from class/method attributes. Hence, `IBackgroundJobClient` just knows nothing about attributes on your method.

shatl:
Awesome! 
Thanks @pieceofsummer

shatl:
Guys, any ETA on release date for this?

shatl:
any news so far?

madannes:
I've just run across this same issue in v1.6.9 on .NET 4.6.2. In glancing at the commits in the pull request, it looks like you're only affecting the .NET Core code. Should a separate issue be created for .NET 4.6.2?

pieceofsummer:
Yes, that was mostly the .NET Core issue.

I have really never used Hangfire on a classic .NET, so can't say for sure. But looking at `Hangfire.AspNet` package, I don't really see any code affecting the job filter providers.

So it might be just your code. Or maybe I'm looking at the wrong package :)

burningice2866:
@pieceofsummer the `Hangfire.AspNet` package only contains some helper-classes to ensure proper shutdown on asp.net - personally i've never used any of it even though i run a dozen of sites using HangFire on Asp.Net.

[...]

PR #800, opened 2017-01-24
Fix issue with configuration block not called on .NET Core + reuse services

pieceofsummer:
On .NET Core the typical initialization of Hangfire is supposed to look like this:
```c#
services.AddHangfire(options => 
{
    options.UseSqlServerStorage("connection string");
    // other things, like options.UseConsole();
});
```
and it seems to work (most of the time).

However, in some cases one might want to move storage initialization away from configuration block, and place it directry into IoC container:
```c#
services.AddSingleton<JobStorage>(serviceProvider =>
{
    var config = serviceProvider.GetRequiredService<IOptions<DatabaseConfiguration>>().Value;
    return new SqlServerStorage(config.ConnectionString, config.Options);
});

services.AddHangfire(options => 
{
    // other things
}
```
And there's a very valid reason for that: you can use a standard `IOptions<T>` (or any other) service for retrieving configuration parameters however you like, which is not possible inside Hangfire's configuration block since `IServiceProvider` is not exposed (and exposing it will likely be a breaking change).

After that, however, the jobs start failing with a `MissingMethodException`: "No parameterless constructor defined for this object". After some investigation, I've found out that Hangfire was now using `SimpleJobActivator` instead of `AspNetCoreJobActivator`, which obviously cannot inject services into constructor. Not only that, it also completely ignored its own configuration block!

After looking into the code, it turned out that initialization of default logger and activator, as well as invocation of the configuration block, was done in the factory method of a JobStorage service. And since the whole service is now replaced, nothing was initialized at all.

It also appears that BackgroundJobServer and Dashboard are using theirs own hard-coded instances of `IBackgroundJobPerformer` & co., even though they are exposed as services, and it should be possible to replace them with a different implementation.

The purpose of this PR is to address all these issues:
- It gets rid of the messy static field that holds the initialization flag.
- It guarantees that the configuration block will be called, even if you would eventually replace all the underlying services. It also guarantees that configuration block will be called in client-only scenarios, given that you're using `IBackgroundJobClient` / `IRecurringJobManager` instances instead of theirs static counterparts. _Who would want to use static wrappers on .NET Core anyway?_
- Background server and Dashboard now also make use of `IBackgroundJobFactory`, `IBackgroundJobStateChanger` and `IBackgroundJobPerformer` IoC services, which were previously used solely by clients. This also eliminates the probable difference between job filters used by client and server counterparts.
- And it all seems like non-breaking changes.

Uhh, I'm tired of writing this text more than from making the PR itself… :)

### DIFF
--- a/src/Hangfire.AspNetCore/HangfireServiceCollectionExtensions.cs
+++ b/src/Hangfire.AspNetCore/HangfireServiceCollectionExtensions.cs
@@ -26,12 +26,12 @@ using Hangfire.States;
 using Microsoft.Extensions.DependencyInjection;
 using Microsoft.Extensions.DependencyInjection.Extensions;
 using Microsoft.Extensions.Logging;
+using Hangfire.Logging;
 
 namespace Hangfire
 {
     public static class HangfireServiceCollectionExtensions
     {
-        private static int _initialized = 0;
 
         public static IServiceCollection AddHangfire(
             [NotNull] this IServiceCollection services,
@@ -39,14 +39,19 @@ namespace Hangfire
         {
             if (services == null) throw new ArgumentNullException(nameof(services));
             if (configuration == null) throw new ArgumentNullException(nameof(configuration));
+            
+            // ===== Configurable services =====
 
-            services.TryAddSingleton(GetInitializedJobStorage);
+            services.TryAddSingletonChecked(_ => JobStorage.Current);
+            services.TryAddSingletonChecked(_ => JobActivator.Current);
+            services.TryAddSingletonChecked(_ => DashboardRoutes.Routes);
+            services.TryAddSingletonChecked<IJobFilterProvider>(_ => JobFilterProviders.Providers);
 
-            services.TryAddSingleton(_ => GlobalConfiguration.Configuration);
-            services.TryAddSingleton(_ => JobActivator.Current);
-            services.TryAddSingleton(_ => DashboardRoutes.Routes);
 
-            services.TryAddSingleton<IJobFilterProvider>(_ => GlobalJobFilters.Filters);
+            // ===== Internal services =====
+
+            // NOTE: these are not required to be checked, because they only depend on already checked configurables,
+            //       are not accessed directly, and can't be affected by customizations made from configuration block.
 
             services.TryAddSingleton<IBackgroundJobFactory>(x => new BackgroundJobFactory(
                 x.GetRequiredService<IJobFilterProvider>()));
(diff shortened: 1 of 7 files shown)


src/Hangfire.AspNetCore/HangfireServiceCollectionExtensions.cs
```diff
@@ -58,62 +63,75 @@ public static class HangfireServiceCollectionExtensions
...
+            // IGlobalConfiguration serves as a marker indicating that Hangfire's services 
+            // were added to the service container (checked by IApplicationBuilder extensions).
+            // 
+            // Being a singleton, it also guarantees that the configuration callback will be 
+            // executed just once upon initialization, so there's no need to double-check that.
+            // 
+            // It should never be replaced by another implementation !!!
+            // AddSingleton() will throw an exception if it was already registered
+
+            services.AddSingleton<IGlobalConfiguration>(serviceProvider =>
```
nickalbrecht: Should this not still be `TryAddSingleton()`, so that if the developer has already specified a custom way to resolve an instance of `IGlobalConfiguration`, you don't blow that out of the water? This is the convention even with Microsoft's `AddMvc()` extension method(s).

src/Hangfire.AspNetCore/HangfireServiceCollectionExtensions.cs
```diff
@@ -58,62 +63,75 @@ public static class HangfireServiceCollectionExtensions
...
-                    var loggerFactory = serviceProvider.GetService<ILoggerFactory>();
-                    if (loggerFactory != null)
-                    {
-                        config.UseLogProvider(new AspNetCoreLogProvider(loggerFactory));
-                    }
-
-                    var scopeFactory = serviceProvider.GetService<IServiceScopeFactory>();
-                    if (scopeFactory != null)
-                    {
-                        config.UseActivator(new AspNetCoreJobActivator(scopeFactory));
```
nickalbrecht: It'd be great to see the calls to instantiate the `LogProvider` and the `JobActivator` instances passed onto the serviceprovider so that they can be overridden

src/Hangfire.AspNetCore/HangfireServiceCollectionExtensions.cs
pieceofsummer: Actually, no, I think. 

`IGlobalConfiguration` is a static instance to call configuration methods on, and it really doesn't make sense in IoC environment. There's no reason for it to be registered in IoC container at all. It only serves as a marker that the services were configured, and as a guarantee that initialization was done once.

If it'd registered with `TryAddSingleton()`, it may be overridden with another implementation, and neither error will happen, nor configuration callback will be fired, which is essentially breaking the purpose of this PR.

src/Hangfire.AspNetCore/HangfireServiceCollectionExtensions.cs
pieceofsummer: They're only initialized to default values before the configuration callback is executed, and they can be overridden later inside the callback.

Even if the default implementation is added here (and overwrites `JobActivator.Current` property), it will still resolve the service from IoC container instead. 

If a custom implementation is registered before the callback is called, it will be used. Otherwise, activator from `JobActivator.Current` will be used (which is either a default implementation defined here, or the result of `UseActivator()` call later in the callback).

src/Hangfire.AspNetCore/HangfireServiceCollectionExtensions.cs
nickalbrecht: Ah, interesting, Makes sense :-)

src/Hangfire.AspNetCore/HangfireServiceCollectionExtensions.cs
nickalbrecht: It seems a waste to instantiate these two classes only to have them changed in the config, but I feel like that just me nit-picking more than anything. I think I'll have to try pulling down this fork and trying it out to see if it solves the problem I was having with the official release. Hopefully the owner of the project can approve your PR

pieceofsummer: These two classes are rather ligtweight, and only instantiated once during application startup. Custom loggers and activators are rarely used on .NET Core, so it's a very small price to pay for not requiring to specify them explicitly every single time.

pieceofsummer:
@odinserj bump

nickalbrecht:
Yeah, I gave up on waiting for this. I ended up cloning `HangfireServiceCollectionExtensions.cs` and tweaking the setup so that I could use my own job activator, and then just do my own scope/unit-of-work resolution inside of that.

gitblit:
Looking forward to these improvements.  I just ran into this today after hanging to the injection approach outlined by @pieceofsummer with manual setting of the static storage singleton.  I've found that I need to use the static registration methods in Asp.Net Core because `IJobClient` does not respect queue attributes.

pieceofsummer:
@gitblit take a look at #799 for a  workaround which works for the current version, until the new one is released.

PR #801, opened 2017-01-24
Ncrontab advanced

shaycraft:
This converts to using Ncrontab-advanced, which enables using more complex cron expressions, which is achieved  by updating from Ncrontab to Ncrontab-advanced.


pieceofsummer:
This PR breaks binary compatibility. You should provide method overloads instead of adding extra argument with a default value.

shaycraft:
Agreed about the default values.  I've now removed this an added additional overloads.

shaycraft:
Can anyone explain to me why the code coverage is failing? I see the following from the report on my changed files:

`src/Hangfire.Core/Server/RecurringJobScheduler.cs	85.54%  no change
src/Hangfire.Core/Server/ScheduleInstant.cs	96.42%	no change
src/Hangfire.Core/RecurringJob.cs	0% no change
src/Hangfire.Core/RecurringJobManagerExtensions.cs	100% (+7.69%)
src/Hangfire.Core/RecurringJobManager.cs	93.65% (0.2%)` - Note also, visuals studio reports coverage on this as different as 95%.`

I must be reading it wrong, but what I see is every file in this pull request either has no change or an increase in coverage, yet the report also says overall coverage will be decreased by 0.41%.  Makes no sense.

shaycraft:
Going to abandon this since this is superseded by Cronos.

PR #832, opened 2017-03-11
Fix number of displayed Records in RecurringJobsPage

reaction1989:
The RecurringJobsPage only displayed 9 instead of 10 items.


odinserj:
Nice catch, thanks a lot @reaction1989 :+1:! Will release the fix today with 1.6.10.

PR #853, opened 2017-03-30
Use Cronos instead of NCrontab

### LINKED ISSUE #494: Adding additional cron support (#, L, W)
This is more of a feature request than an issue.  We needed Hangfire to support a few cron features that the current version didn't support (# - day of week field, L - last day of week/last day of month, W - closest weekday to specified day.)

After digging through NCrontab's code this weekend, I determined it wouldn't be easy to implement within the current codebase, so I re-wrote the parser and created a new project called NCrontab-Advanced.

I'm currently testing a branch of Hangfire with these changes in place.  Just putting this issue here as a placeholder in case anyone else needs this functionality.  As soon as I have enough unit tests added to NCrontab-Advanced that I feel confident it's ready for prime time, I'll initiate the pull request.

Repos:
[Hangfire using NCrontab-Advanced](https://github.com/jcoutch/Hangfire)
[NCrontab-Advanced repo](https://github.com/jcoutch/NCrontab-Advanced)

aidmsu:
This PR fixes #494, #567 by using **[Cronos](https://github.com/HangfireIO/Cronos)** instead of **NCrontab** in `RecurringJobScheduler`.

Pros of using **Cronos**:

* **time zones** support out of the box;
* correct **Daylight Saving Time** handling;
* support of non-standard characters: **L**, **W**, **#**, **?**;
* **performance**: it's really very fast.

It means that:
* recurring jobs **won't be missed** on standard time (ST) to **[Daylight Saving Time](https://en.wikipedia.org/wiki/Daylight_saving_time)** (DST) transitions (when the clock jumps forward);
* minutely, hourly jobs **won't be missed** on DST to ST transitions (when the clock jumps backward);
* daily, monthly jobs **won't be repeated** on DST to ST transitions;
* users can schedule recurring jobs more flexible using non-standard characters. Example for **last weekday of month**: ```RecurringJob.AddOrUpdate(() => SomeMethod(), "* * LW * *")```

Read [more](https://github.com/HangfireIO/Cronos#cronos) about all features.


odinserj:
Finally we'll get around all the problems with Cron. Almost all, seconds still aren't supported yet 😉

patnolan:
Can you provide any guidance around when seconds will be supported? 

I'm trying to make a decision between Quartz.net and Hangfire but the main requirement is the support for a trigger with a 45sec interval.

tanushshukla:
Looks like 1.7.0 beta build is indeed using Cronos. 

Thanks a lot! I'll be testing it out.

PR #858, opened 2017-04-05
Dnx

gmahota:


pieceofsummer:
What's the point of migration to an unsupported project format? And targeting a rather small subset of environments at that?

PR #869, opened 2017-04-18
Change calculation for AutomaticRetryAttribute.cs SecondsToDelay

Adrian-Nelson:
The current implementation of SecondsToDelay uses "15 + random.Next(30)" which will generate potential collisions between subsequent requests made within the 30 seconds of each other - so this can lead to higher spikes of retry in a pathological scenario.

For example, if you have 26 failing items (A to Z) which occur at time T, (T + 1), ... , (T + 25) it is possible that all 26 would be scheduled for a retry at the same time in the future because item A may have a random value of 25, B a random value of 24, C a random value of 23 etc etc (unlikely but possible).

Obviously if you have a lot of items retrying at exactly the same time the retry load will be the same in the future with a constant factor but it will be predictable.

But with a random element in play you could end up with randomly much higher retry processing spikes.

My suggestion would be to simplify and use a constant prime number such as "43" and this way the retries will be more deterministic and not produce potentially higher spikes of work.

Note: This solution also eliminates the need to allocate and dispose a new PRNG with each retry attempt.

### DIFF
--- a/src/Hangfire.Core/AutomaticRetryAttribute.cs
+++ b/src/Hangfire.Core/AutomaticRetryAttribute.cs
@@ -249,9 +249,8 @@ namespace Hangfire
         // delayed_job uses the same basic formula
         private static int SecondsToDelay(long retryCount)
         {
-            var random = new Random();
             return (int)Math.Round(
-                Math.Pow(retryCount - 1, 4) + 15 + random.Next(30) * retryCount);
+                Math.Pow(retryCount - 1, 4) + 43 * retryCount);
         }
     }
 }


odinserj:
Hm, but what if all of them, 26 items, are failing at the same time T (unlikely, but possible too)? With the new implementation, all of them will be scheduled in T1 for the first attempt, T2 for the second attempt and so on. Randomization at least provides an opportunity to be scheduled at different times for next retries.

Adrian-Nelson:
But if all 26 failed and were re-scheduled this would be no more load than the original 26 failures.

The issue I have with the randomization is that it could lead to much higher loads as it would be possible that a lot of failures get retried at the same time which could lead to higher spikes in processing load which is what I would like to see eliminated.

I realize that this is an edge case but if anything is processing large volumes then the randomization of retries could make itself apparent with higher than expected processing spikes - I hope you see my (hypothetical) concern here.

If you feel strongly that the randomization is fine and that no one has reported unexpected processing spikes in real life then please do ignore this... although I've found from experience on other projects that randomization of retry behavior can lead to unpredictable and hard to reproduce run-time issues.

odinserj:
> But if all 26 failed and were re-scheduled this would be no more load than the original 26 failures.

Only if there's only one retry attempt, and only if jobs are failed at different times, fixed-time retries are better. But if they are failed due to an upstream fault, they will be likely fail at the same time. In this case it will be better to attempt to spread the load with a bit of randomness.

This is a well-established practice, don't see any important reason yet to change this behavior. PR #931 will allow to change the default retry delay policy.

Adrian-Nelson:
OK - I agree in that sort of situation it would be better to try and spread the load.

As you state the other PR will allow this to be changed by the application consuming HangFire to allow it to control the retries in a way that best suits that particular application.

So I guess this PR can be ignored / closed. Thanks for considering it..

odinserj:
Thanks for sharing your thoughts, @Adrian-Nelson! It's always good to re-consider the choices to have the best default behaviour. Yes, that PR will allow to have a custom logic to specify delays for retries, so we definitely can close the issue.

PR #887, opened 2017-05-17
AspNetCoreJobActivatorScope Resolve using ActivatorUtilities.GetServiceOrCreateInstance

Tsabo:
Use ActivatorUtilities.GetServiceOrCreateInstance instead of ServiceProvider.GetRequiredService, this allows creating instances of notification handlers without requiring them to be registered in DoI. Otherwise GetRequiredService would throw an error if a service isn't registered. This allow the handler to be created and still apply constructor injection to the handler.


odinserj:
Indeed, good catch, @Tsabo! Will be available in Hangfire 1.6.13.

PR #891, opened 2017-05-19
Reduce serialization size

### LINKED ISSUE #816: Enqueue from .NET Core, execute on classic .NET
There's a cross-framework type name serialization issue. 

Common built-in types (like String, Int32 etc.), which were previously defined in `mscorlib`, are now sitting in `System.Private.CoreLib` on .NET Core, so the classic .NET framework will fail to deserialize them because of a missing assembly. The problem is actually one-sided, because .NET Core has type redirection and will successfully deserialize mscorlib types.

It sure can be workarounded by providing a custom `JsonConverter` for JobHelper, but it would be nice for this to be handled out-of-the-box.

For reference, the converter itself might be like this:
```c#
public class CompatibleTypeConverter : JsonConverter
{
    public override bool CanConvert(Type objectType) => typeof(Type).IsAssignableFrom(objectType);

    public override object ReadJson(JsonReader reader, Type objectType, object existingValue, JsonSerializer serializer)
    {
        switch (reader.TokenType)
        {
            case JsonToken.Null:
                return null;

            case JsonToken.String:
                return Type.GetType((string)reader.Value, true);

            default:
                throw new JsonSerializationException($"Unexpected token when reading type name: {reader.TokenType}");
        }
    }

    public override void WriteJson(JsonWriter writer, object value, JsonSerializer serializer)
    {
        if (value == null)
        {
            writer.WriteNull();
            return;
        }

        var type = (Type)value;
        var typeName = type.AssemblyQualifiedName;

#if NETSTANDARD1_3 || NETSTANDARD1_4 || NETSTANDARD1_5 || NETSTANDARD1_6 || NETCOREAPP1_0 || NETCOREAPP1_1 // || ...
        var assemblyName = type.GetTypeInfo().Assembly.GetName();
        if (assemblyName.Name == "System.Private.CoreLib")
        {
            // AssemblyName is copied, so we can overwrite Name in-place
            assemblyName.Name = "mscorlib";

            var mscorlibTypeName = Assembly.CreateQualifiedName(assemblyName.FullName, type.FullName);

            if (type == Type.GetType(mscorlibTypeName, false))
            {
                // type is redirected
                typeName = mscorlibTypeName;
            }
        }
#endif

        writer.WriteValue(typeName);
    }
}
```

aidmsu:
This PR fixes #816.

Serialization is cumbersome in some places. I propose the following changes:

### Job serialization
- Do not save redundant assembly's information: Version, Culture, PublicKeyToken;
- Do not save assembly names for system types: "mscorlib" for .NET Full Framework and "System.Private.Core" for .NET Core
- Serialize InvocationData as properties array instead of json object;

**Example:**

Consider the following code
```
BackgroundJob.Enque(() => Console.WriteLine("Hello!"));
```
Current serialization of InvocationData looks like this:
```
"{\"Type\":\"System.Console, mscorlib, Version=4.0.0.0, Culture=neutral, PublicKeyToken=b77a5c561934e089\",\"Method\":\"WriteLine\",\"ParameterTypes\":\"[\\\"System.String, mscorlib, Version=4.0.0.0, Culture=neutral, PublicKeyToken=b77a5c561934e089\\\"]\",\"Arguments\":\"[\\\"\\\\\\\"Hello!\\\\\\\"\\\"]\"}")]
```
New format serialization looks like this:
```
["System.Console","WriteLine","[\"System.String\"]","[\"\\\\\"Hello!\\\\\"\"]"]
```
### State serialization
- Mark some state properties (like `IsFinal`, `IgnoreJobLoadException`) by `JsonIgnore` attribute.
- Do not save the default values when `AwaitingState.NextState` property is serialized.
- Sarialize `AwatingState.Options` as integer instead of string.

**Example:**

Current `AwatingState` serialization format:
```
{"ParentId":"1","NextState":"{\"$type\":\"Hangfire.States.EnqueuedState, Hangfire.Core\",\"Queue\":\"default\",\"EnqueuedAt\":\"2017-05-19T10:59:23.0050945Z\",\"Name\":\"Enqueued\",\"Reason\":null,\"IsFinal\":false,\"IgnoreJobLoadException\":false}","Options":"OnlyOnSucceededState","Expiration":"365.00:00:00"}
```
New format:
```
{"ParentId":"1","NextState":"{\"$type\":\"Hangfire.States.EnqueuedState, Hangfire.Core\",\"EnqueuedAt\":\"2017-05-19T10:59:23.0050945Z\"}","Options":1,"Expiration":"365.00:00:00"}
```

### DIFF
--- a/src/Hangfire.Core/Storage/InvocationData.cs
+++ b/src/Hangfire.Core/Storage/InvocationData.cs
@@ -74,9 +74,9 @@ namespace Hangfire.Storage
         public static InvocationData Serialize(Job job)
         {
             return new InvocationData(
-                SerializeType(job.Type),
+                SerializeType(job.Type).ToString(),
                 job.Method.Name,
-                SerializeTypes(job.Method.GetParameters().Select(x => x.ParameterType).ToArray()),
+                SerializeTypes(job.Method.GetParameters().Select(x => x.ParameterType).ToArray()).ToString(),
                 JobHelper.ToJson(SerializeArguments(job.Args)));
         }
 
@@ -251,46 +251,32 @@ namespace Hangfire.Storage
             return result;
         }
 
-        private static string SerializeTypes(IList<Type> types)
+        private static StringBuilder SerializeTypes(Type[] types, char beginTypeDelimiter = '"', char endTypeDelimiter = '"', StringBuilder typeNamesBuilder = null)
         {
             if (types == null) return null;
+            if (typeNamesBuilder == null) typeNamesBuilder = new StringBuilder();
 
-            var typeNamesBuilder = new StringBuilder();
-            SerializeTypes(typeNamesBuilder, types, '"', '"');
-
-            return typeNamesBuilder.ToString();
-        }
-
-        private static string SerializeType(Type type)
-        {
-            var typeNameBuilder = new StringBuilder();
-            SerializeType(typeNameBuilder, type);
-
-            return typeNameBuilder.ToString();
-        }
-
-        private static void SerializeTypes(StringBuilder typeNamesBuilder, IList<Type> types, char beginTypeDelimiter, char endTypeDelimiter)
-        {
             typeNamesBuilder.Append('[');
             
-            for (var i = 0; i < types.Count; i++)
+            for (var i = 0; i < types.Length; i++)
             {
                 typeNamesBuilder.Append(beginTypeDelimiter);
-                SerializeType(typeNamesBuilder, types[i]);
+                SerializeType(types[i], true, typeNamesBuilder);
                 typeNamesBuilder.Append(endTypeDelimiter);
 
-                if (i != types.Count - 1) typeNamesBuilder.Append(',');
+                if (i != types.Length - 1) typeNamesBuilder.Append(',');
             }
-
-            typeNamesBuilder.Append(']');
+            
+            return typeNamesBuilder.Append(']');
         }
 
-        private static void SerializeType(StringBuilder typeNameBuilder, Type type, bool withAssemblyName = true)
+        private static StringBuilder SerializeType(Type type, bool withAssemblyName = true, StringBuilder typeNameBuilder = null)
         {
+            typeNameBuilder = typeNameBuilder ?? new StringBuilder();
+
             if (type.DeclaringType != null)
             {
-                SerializeType(typeNameBuilder, type.DeclaringType, false);
-                typeNameBuilder.Append('+');
+                SerializeType(type.DeclaringType, false, typeNameBuilder).Append('+');
             }
             else if (type.Namespace != null)
             {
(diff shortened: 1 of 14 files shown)


burningice2866:
How will this affect existing Storage Providers which will keep using `JobHelper.FromJson<InvocationData>` until they eventually get updated. I assume this line will make sure things keep working https://github.com/HangfireIO/Hangfire/pull/891/files#diff-762b9c9bc1acba9c001d8544882b1d4aR85?

aidmsu:
Hi, @burningice2866 ! 

Thank you for good question.

1. This PR doesn't affect existing Storage Providers. Storage Provider itself is responsible for serialization and deserialization jobs. It means that if  Storage Provider use `JobHelper.FromJson<InvocationData>` for `InvocationData` deserialization it uses `JobHelper.ToJson` for `InvocationData` serialization.
So these changes affect only `HangFire.SqlServer` provider (see [changes](https://github.com/HangfireIO/Hangfire/pull/891/files#diff-808469daa5e53631edd5e3912b6d2987L93) in "src/Hangfire.SqlServer" folder).
2. These changes affect recurring jobs serialization regardless of Storage Provider.
3. Yes, you are right, [this line](https://github.com/HangfireIO/Hangfire/pull/891/files#diff-762b9c9bc1acba9c001d8544882b1d4aR85) was added to support existing jobs and recurring jobs which are stored in previous format.

src/Hangfire.Core/Storage/InvocationData.cs
```diff
@@ -224,5 +250,48 @@ internal static bool ParseDateTimeArgument(string argument, out DateTime value)
...
+                var serializedTypeArgs = string.Join(",", type.GenericTypeArguments.Select(typeArg =>
+                {
+                    var serializedType = SerializeType(typeArg);
+                    return $"[{serializedType}]";
+                }));
+
+                typeName += $"[{serializedTypeArgs}]";
+            }
+
+            var assemblyName = type.GetTypeInfo().Assembly.GetName().Name;
```
pieceofsummer: If `withAssemblyName == false`, can just return `typeName` here and avoid extra reflection and allocation of `AssemblyName`, no?

pieceofsummer:
I'd also suggest using `StringBuilder` when constructing type names instead of string concatenations and LINQ. All that syntactic sugar is nice and everything, but serialization is quite a performance-critical part, and avoiding extra memory allocations is good.

aidmsu:
@pieceofsummer thanks for review!

I rewrote type serialization using `StringBuilder`.

src/Hangfire.Core/Storage/InvocationData.cs
```diff
@@ -224,5 +250,68 @@ internal static bool ParseDateTimeArgument(string argument, out DateTime value)
             value = dateTime;
             return result;
         }
+
+        private static string SerializeTypes(IList<Type> types)
```
odinserj: Since the type of the only argument that's passed is `Type[]`, you can safely change it to array to get rid of unneeded interface.

odinserj:
@aidmsu, did you remember we were considering how to reduce the number of backslashes in argument serialisation to avoid `"[\"\\\\\"Hello!\\\\\"\"]"]"` and maybe `"[\"System.String\"]"`? What was our decision on that?

aidmsu:
@odinserj I'm investigating how we can reduce the number of backslashes now.

aidmsu:
@odinserj Finally it's possible to reduce the nubmer of backslashes and we can serialize `InvocationData` to something like `["System.Console","WriteLine",["System.String"],["\\\"Hello!\\\"]]`. But for a sake of multilevel deserialization: `string - InvocationData -> Job` it will be difficult to implement and maintain in the future.

odinserj:
LGTM, merging

Issue #893, opened 2017-05-20
Generic Methods that don't have all of the generic types in the argument list cannot be enqueued


sage-nathan-jones:
So from the test suite, the method `GenericMethod<T>() {}` is never used in the tests. The current behavior is to throw a Null Pointer Exception when trying to enqueue.

This test demonstrates the issue:
```
        [Fact]
        public void GetNonOpenMatchingMethod_HandlesMethodHasNoParametersOrTypes()
        {
            var method = TypeExtensions.GetNonOpenMatchingMethod(typeof(NonGenericClass), "GenericMethod",
                new Type[0]);
            Assert.Null(method);
        }
```
An exception is thrown rather than the Assert.Null()

sage-nathan-jones:
#894 is a PR that addresses this and similar issues where the generic types are not in the method's arguments.

odinserj:
Thanks a lot for PR, @sage-nathan-jones, especially for backward compatibility! We'll review it a bit later, with combination of PR #891.

odinserj:
I've added the proposed tests to ensure we aren't getting an exception if a class contains such methods, and just ignore methods that contain unresolved type arguments. Full fix requires us to serialize type arguments into the method name like `MyMethod[[System.String, mscorlib],[System.Int32, mscorlib]]` and require much more efforts, especially when thinking about backward compatibility.

PR #894, opened 2017-05-20
add support for generic methods where the generic types may not be in…

sage-nathan-jones:
… the method's argument list. Maintains backward compatibility with jobs already serialized before this change.

Addresses issue #893

### DIFF
--- a/src/Hangfire.Core/Common/TypeExtensions.cs
+++ b/src/Hangfire.Core/Common/TypeExtensions.cs
@@ -42,7 +42,8 @@ namespace Hangfire.Common
         public static MethodInfo GetNonOpenMatchingMethod(
             [NotNull] this Type type,
             [NotNull] string name,
-            [CanBeNull] Type[] parameterTypes)
+            [CanBeNull] Type[] parameterTypes,
+            [CanBeNull] Type[] genericTypes = null)
         {
             if (type == null) throw new ArgumentNullException(nameof(type));
             if (name == null) throw new ArgumentNullException(nameof(name));
@@ -70,12 +71,20 @@ namespace Hangfire.Common
                     continue;
                 }
 
-                var parameterTypesMatched = true;
 
                 var genericArguments = methodCandidate.ContainsGenericParameters
-                    ? new Type[methodCandidate.GetGenericArguments().Length]
-                    : null;
-                
+                        ? new Type[methodCandidate.GetGenericArguments().Length]
+                        : null;
+
+                // Determine if it is even possible for these signatures to match based on generics
+                if (genericArguments != null && genericTypes != null)
+                {
+                    if (genericTypes.Length != genericArguments.Length)
+                        continue;
+                }
+
+                var parameterTypesMatched = true;
+
                 // Determining whether we can use this method candidate with
                 // current parameter types.
                 for (var i = 0; i < parameters.Length; i++)
@@ -92,9 +101,27 @@ namespace Hangfire.Common
 
                 if (parameterTypesMatched)
                 {
+                    if (genericTypes == null || genericTypes.Length == 0)
+                    {
+                        // existing behavior using inference from arguments
+                        genericTypes = genericArguments;
+                    }
+
+                    if (genericTypes != null)
+                    {
+                        // Determine if inferred and specified match
+                        parameterTypesMatched = genericTypes.Zip(genericArguments, (l, r) => r == null || l == r).Aggregate(true, (acc, res) => acc && res);
+                        // last check, Ensure that all genericTypes have been specified
+                        // (this prevents a NullPointerException below)
+                        parameterTypesMatched &= genericTypes.Aggregate(true, (acc, res) => acc && res != null);
+                    }
+                }
+
+                if (parameterTypesMatched)
+                {
                     // Return first found method candidate with matching parameters.
-                    return genericArguments != null
-                        ? methodCandidate.MakeGenericMethod(genericArguments)
+                    return (genericTypes != null && genericTypes.Length > 0)
+                        ? methodCandidate.MakeGenericMethod(genericTypes)
                         : methodCandidate;
                 }
             }
--- a/src/Hangfire.Core/Storage/InvocationData.cs
+++ b/src/Hangfire.Core/Storage/InvocationData.cs
@@ -30,17 +30,19 @@ namespace Hangfire.Storage
     public class InvocationData
     {
         public InvocationData(
-            string type, string method, string parameterTypes, string arguments)
+            string type, string method, string parameterTypes, string arguments, string genericTypes = null)
         {
             Type = type;
             Method = method;
             ParameterTypes = parameterTypes;
             Arguments = arguments;
+            GenericTypes = genericTypes;
         }
 
         public string Type { get; }
         public string Method { get; }
         public string ParameterTypes { get; }
+        public string GenericTypes { get; }
         public string Arguments { get; set; }
 
         public Job Deserialize()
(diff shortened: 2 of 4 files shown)


src/Hangfire.Core/Storage/InvocationData.cs
```diff
@@ -30,17 +30,19 @@ namespace Hangfire.Storage
...
             Method = method;
             ParameterTypes = parameterTypes;
             Arguments = arguments;
+            GenericTypes = genericTypes;
         }
 
         public string Type { get; }
         public string Method { get; }
         public string ParameterTypes { get; }
+        public string GenericTypes { get; }
```
aidmsu: Consider not to create separate property `GenericTypes`. Can we store info about generic arguments in `Method` property? `Method` value may look like this:
```
Generic[\"System.Int32\"]
```

src/Hangfire.Core/Common/TypeExtensions.cs
```diff
@@ -70,12 +71,20 @@ public static string ToGenericTypeString(this Type type)
                     continue;
                 }
 
-                var parameterTypesMatched = true;
 
                 var genericArguments = methodCandidate.ContainsGenericParameters
-                    ? new Type[methodCandidate.GetGenericArguments().Length]
-                    : null;
-                
+                        ? new Type[methodCandidate.GetGenericArguments().Length]
```
aidmsu: Could we rename `genericTypes` method parameter to `genericArguments`? It will be able to simplify code. The onle change will be in `genericArguments` assignment:
```
if (methodCandidate.ContainsGenericParameters)
{
    var genericArgumentsCount = methodCandidate.GetGenericArguments().Length;
    if (genericArguments == null) genericArguments = new Type[genericArgumentsCount];
    if (genericArgumentsCount != genericArguments.Length) continue;
}

File: src/Hangfire.Core/Storage/InvocationData.cs (hunk 1 above)
Comment (sage-nathan-jones) [inline src/Hangfire.Core/Storage/InvocationData.cs]: I suppose that it could be done as you suggest, but if I understand the code, then the serialization and deserialization of the InvocationData would become more convoluted. Instead of just parsing a string (method) and JSON array (types) from separate strings in the JSON object, we'd have first take apart the .Method and then deserialize the components separately.

File: src/Hangfire.Core/Common/TypeExtensions.cs (hunk 2 above)
Comment (sage-nathan-jones) [inline src/Hangfire.Core/Common/TypeExtensions.cs]: I don't think that I understand your suggestion, so I will present what my intents are.
Currently hangfire looks for a generic method signature by looking for the number of the methods generic types, but it populates these types by looking at the generic arguments. It cannot handle the case where there are types not in the argument list. (call this item "A") The loop at lines 79-90 are responsible for by calling TypesMatchRecursive which overwrites the values passed into the method's genericArguments variable. (item "B")

There is a bug in hangfire where if you try to use a generic method that has types which are not represented in the argument list of the method,(item "A" above) you get a NullReferenceException from the job invocation which is not helpful in resolving the matter. (item "C")

There seems to be no way to address this except for to include all of the generic types from the MethodInfo for serialization. (Hence the changes in InvocationData.) Once this metadata is included, then there is no need to "sniff" the generic types by looking at the argument list. (item "D")

So this PR tries to do multiple things: 

1) fix the issue by serializing all of the data needed and then using this info for invocation. Since this is taken from MethodInfo it should be considered correct over any other methods. 

2) It allows job serialized without the generic type metadata to still work within the constraints that are currently in place (namely arguments contain at least one of each generic type.) However, this sniffing will not be considered correct over explicit info (item "B" above). 

3) It changes the behavior of any pre-existing job that does not follow the constraint such that it does not throw a NullReferenceException in the job invocation, rather it behaves as if the method was not found (which is more accurate IMO). (item "C" above)

So I chose to leave the existing sniffing alone for maximum backwards compat. But only look at the sniffing in the case of missing explicit types. And even then I make sure that they're all there to prevent the unhelpful NullPointerException that we were originally experiencing.

With this context, could you rephrase your comment?

**Comment by sage-nathan-jones** [comment]
Ok, so I renamed the variables and moved around the logic as you suggested. I removed the code that was in place to check genericTypes against genericArguments since collapsing them to the same variable prohibits that (this was the .Zip() call.) This was a sanity test in any event. 

I also changed the .Aggregate() call to just be a foreach loop. This is still required in my opinion because without it, you will get the NullReferenceException during the MakeGenericMethod() call (which was what our original problem was.) This check has to be done after any sort of sniffing from TypeMatchRecursive().

**Comment by sage-nathan-jones** [comment]
I've made the requested changes. It has the original behavior of throwing a NullReferenceObject from MethodInfo.MakeGeneric()

**Comment by odinserj** [comment]
@sage-nathan-jones thanks a lot for your hard work! @aidmsu, to what branch should we merge these changes?

File: src/Hangfire.Core/Storage/InvocationData.cs [hunk 3]
```diff
@@ -30,17 +30,19 @@ namespace Hangfire.Storage
     public class InvocationData
     {
         public InvocationData(
-            string type, string method, string parameterTypes, string arguments)
```
aidmsu: I found one more problem. If we just add optional parameter `string genericTypes = null` to `InvocationData` constructor Hangfire.Core can become binary incompatible with Hangfire storage libraries such as Hangfire.Pro.Redis.

Instead I suggest to overload existing constructor and mark it by `JsonConstructorAttribute`:
```csharp
[JsonConstructor]
public InvocationData(string type, string method, string parameterTypes, string arguments, string genericTypes)
{
    Type = type;
    Method = method;
    ParameterTypes = parameterTypes;
    Arguments = arguments;
    GenericTypes = genericTypes;
}
```

aidmsu:
@odinserj @sage-nathan-jones I found one more problem in the PR. Since this is bugfix I think we should merge these changes into master after fixing it.

src/Hangfire.Core/Storage/InvocationData.cs
aidmsu: Ah, I missed an important thing(

If we add new `GenericTypes` property and create appropriate constructror `InvocationData(string type, string method, string parameterTypes, string arguments, string genericTypes)` we won't fix the bug for users who use [HangFire.Redis.StackExchange](https://github.com/marcoCasamento/Hangfire.Redis.StackExchange) or [Hangfire.Pro.Redis](https://nuget.hangfire.io/feeds/hangfire-pro/Hangfire.Pro.Redis/) for example. 

These packages don't use `JobHelper.ToJson` to serialize `InvocationData` to string and it won't store `GenericTypes` property automatically. Moreover in order to fix the bug for such users the new version of HangFire.Pro.Redis will have to depend on Hangfire.Core last version instead of Hangfire.Core 1.4.0. 

Therefore I suggest not to create separate property `GenericTypes` and store info about method generic parameters in `Method` property.

odinserj:
@sage-nathan-jones, how are you doing? We realised that fix for this issue is a bit more complicated, because some storages (for example `Hangfire.Redis.StackExchange`) are using properties of the `InvocationData` class directly, without calls to `Serialize` or `Deserialize`. Therefore we can't add any properties here.

As a workaround, we can include the information regarding generic arguments to the `Method` property itself, for example `MyGenericMethod<System.String, System.Int32>`, but this will require additional parsing and much more effort for backward compatibility. 

I would like to know if you still want to complete this issue. If not – not a problem!

sage-nathan-jones:
I will look at it this week. The simple cases should be straightforward, I worry about robustness... but since it is just flat out broken now, I guess it will be better than nothing.

Get Outlook for iOS<https://aka.ms/o0ukef>
________________________________

PR #901, opened 2017-05-26
Change the order of the AutomaticRetryAttribute filter

odinserj:
This PR increases the order of the `AutomaticRetryAttribute` filter to `100`, to be able to pay more attention to exceptions that are thrown just before an automatic retry. This change means the following:

1. It will allow to intercept transition to the `FailedState` before running `AutomaticRetryAttribute`'s logic, for example, to allow logging all the retries.
2. It will be performed *after* the `StatisticsHistory` filter that tracks number of failed jobs that's shown in the dashboard.


odinserj:
Hmm...

PR #903, opened 2017-05-29
Change `AutomaticRetryAttribute` filter Order to 20

odinserj:
This change will allow to add elect state filters that are able intercept the `FailedState` transition before the `AutomaticRetryAttribute` filter. For example, to set custom notifications about retries, provide custom automatic retry logic, etc. Currently this isn't possible without clearing the global filters, because `AutomaticRetryAttribute` has the lowest possible order.


dcorbett-fivebridgesllc:
+1  This change will be very helpful...

PR #912, opened 2017-06-05
JobStorage has a prop to define job expiration timeout

rsilva93:
To make it possible to configure expiration timeout for jobs :)


rsilva93:
Need a fix.

PR #913, opened 2017-06-05
JobStorage has a prop to define job expiration timeout

rsilvanet:
Hello,

The idea is make it possible to configure expiration timeout for jobs :)

Best regards.

### DIFF
--- a/src/Hangfire.Core/JobStorage.cs
+++ b/src/Hangfire.Core/JobStorage.cs
@@ -54,7 +54,7 @@ namespace Hangfire
             }
         }
 
-        public virtual TimeSpan JobExpirationTimeout
+        public TimeSpan JobExpirationTimeout
         {
             get
             {
--- a/src/Hangfire.Core/JobStorage.cs
+++ b/src/Hangfire.Core/JobStorage.cs
@@ -58,7 +58,10 @@ namespace Hangfire
         {
             get
             {
-                return _jobExpirationTimeout;
+                lock (LockObject)
+                {
+                    return _jobExpirationTimeout;
+                }
             }
             set
             {
--- a/src/Hangfire.Core/JobStorage.cs
+++ b/src/Hangfire.Core/JobStorage.cs
@@ -28,7 +28,8 @@ namespace Hangfire
     {
         private static readonly object LockObject = new object();
         private static JobStorage _current;
-        private static TimeSpan _jobExpirationTimeout = TimeSpan.FromDays(1);
+
+        private TimeSpan _jobExpirationTimeout = TimeSpan.FromDays(1);
 
         public static JobStorage Current
         {
--- a/src/Hangfire.Core/JobStorage.cs
+++ b/src/Hangfire.Core/JobStorage.cs
@@ -28,6 +28,7 @@ namespace Hangfire
     {
         private static readonly object LockObject = new object();
         private static JobStorage _current;
+        private static TimeSpan _jobExpirationTimeout = TimeSpan.FromDays(1);
 
         public static JobStorage Current
         {
@@ -52,8 +53,23 @@ namespace Hangfire
             }
         }
 
+        public virtual TimeSpan JobExpirationTimeout
+        {
+            get
+            {
+                return _jobExpirationTimeout;
+            }
+            set
+            {
+                lock (LockObject)
+                {
+                    _jobExpirationTimeout = value;
+                }
+            }
+        }
+
         public abstract IMonitoringApi GetMonitoringApi();
-        
+
         public abstract IStorageConnection GetConnection();
 
 #pragma warning disable 618
--- a/src/Hangfire.Core/JobStorage.cs
+++ b/src/Hangfire.Core/JobStorage.cs
@@ -73,7 +73,7 @@ namespace Hangfire
         }
 
         public abstract IMonitoringApi GetMonitoringApi();
-
+        
         public abstract IStorageConnection GetConnection();
 
 #pragma warning disable 618
--- a/src/Hangfire.Core/JobStorage.cs
+++ b/src/Hangfire.Core/JobStorage.cs
@@ -65,6 +65,11 @@ namespace Hangfire
             }
             set
             {
+                if (value < TimeSpan.FromHours(1))
+                {
+                    throw new ArgumentException("JobStorage.JobExpirationTimeout value should be equal or greater than 1 hour.");
+                }
+
                 lock (LockObject)
                 {
                     _jobExpirationTimeout = value;
@@ -73,7 +78,7 @@ namespace Hangfire
         }
 
         public abstract IMonitoringApi GetMonitoringApi();
-        
+
         public abstract IStorageConnection GetConnection();
 
 #pragma warning disable 618
--- a/src/Hangfire.Core/JobStorage.cs
+++ b/src/Hangfire.Core/JobStorage.cs
@@ -58,10 +58,7 @@ namespace Hangfire
         {
             get
             {
-                lock (LockObject)
-                {
-                    return _jobExpirationTimeout;
-                }
+                return _jobExpirationTimeout;
             }
             set
             {
@@ -70,10 +67,7 @@ namespace Hangfire
                     throw new ArgumentException("JobStorage.JobExpirationTimeout value should be equal or greater than 1 hour.");
                 }
 
-                lock (LockObject)
-                {
-                    _jobExpirationTimeout = value;
-                }
+                _jobExpirationTimeout = value;
             }
         }
 
--- a/src/Hangfire.Core/States/ApplyStateContext.cs
+++ b/src/Hangfire.Core/States/ApplyStateContext.cs
@@ -25,7 +25,7 @@ namespace Hangfire.States
 #pragma warning restore 618
     {
         public ApplyStateContext(
-            [NotNull] IWriteOnlyTransaction transaction,
+            [NotNull] IWriteOnlyTransaction transaction, 
             [NotNull] ElectStateContext context)
             : this(context.Storage, context.Connection, transaction, context.BackgroundJob, context.CandidateState, context.CurrentState)
         {
@@ -36,7 +36,7 @@ namespace Hangfire.States
             [NotNull] IStorageConnection connection,
             [NotNull] IWriteOnlyTransaction transaction,
             [NotNull] BackgroundJob backgroundJob,
-            [NotNull] IState newState,
+            [NotNull] IState newState, 
             [CanBeNull] string oldStateName)
         {
             if (storage == null) throw new ArgumentNullException(nameof(storage));
@@ -44,7 +44,7 @@ namespace Hangfire.States
             if (transaction == null) throw new ArgumentNullException(nameof(transaction));
             if (backgroundJob == null) throw new ArgumentNullException(nameof(backgroundJob));
             if (newState == null) throw new ArgumentNullException(nameof(newState));
-
+            
             BackgroundJob = backgroundJob;
 
             Storage = storage;
@@ -63,7 +63,7 @@ namespace Hangfire.States
 
         [NotNull]
         public IWriteOnlyTransaction Transaction { get; }
-
+        
         public override BackgroundJob BackgroundJob { get; }
 
         [CanBeNull]
@@ -71,7 +71,7 @@ namespace Hangfire.States
 
         [NotNull]
         public IState NewState { get; }
-
+        
         public TimeSpan JobExpirationTimeout { get; set; }
     }
 }
--- a/src/Hangfire.Core/States/ApplyStateContext.cs
+++ b/src/Hangfire.Core/States/ApplyStateContext.cs
@@ -25,7 +25,7 @@ namespace Hangfire.States
 #pragma warning restore 618
     {
         public ApplyStateContext(
-            [NotNull] IWriteOnlyTransaction transaction, 
+            [NotNull] IWriteOnlyTransaction transaction,
             [NotNull] ElectStateContext context)
             : this(context.Storage, context.Connection, transaction, context.BackgroundJob, context.CandidateState, context.CurrentState)
         {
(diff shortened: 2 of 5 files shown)


rsilva93:
@odinserj :)

src/Hangfire.Core/JobStorage.cs
```diff
@@ -52,8 +54,23 @@ public static JobStorage Current
             }
         }
 
+        public virtual TimeSpan JobExpirationTimeout
+        {
+            get
+            {
+                return _jobExpirationTimeout;
```
odinserj: This line should be locked too.

src/Hangfire.Core/JobStorage.cs
```diff
@@ -52,8 +54,23 @@ public static JobStorage Current
...
+                lock (LockObject)
+                {
+                    _jobExpirationTimeout = value;
+                }
+            }
+        }
+
         public abstract IMonitoringApi GetMonitoringApi();
-        
+
```
odinserj: Lines without meaningful changes shouldn't be committed.

src/Hangfire.Core/States/ApplyStateContext.cs
```diff
@@ -36,23 +36,23 @@ public class ApplyStateContext : StateContext
...
+            [NotNull] IState newState,
             [CanBeNull] string oldStateName)
         {
             if (storage == null) throw new ArgumentNullException(nameof(storage));
             if (connection == null) throw new ArgumentNullException(nameof(connection));
             if (transaction == null) throw new ArgumentNullException(nameof(transaction));
             if (backgroundJob == null) throw new ArgumentNullException(nameof(backgroundJob));
             if (newState == null) throw new ArgumentNullException(nameof(newState));
-            
+
```
odinserj: 5 changed lines in this class without meaningful changes, should be removed.

src/Hangfire.Core/JobStorage.cs
```diff
@@ -52,8 +54,23 @@ public static JobStorage Current
             }
         }
 
+        public virtual TimeSpan JobExpirationTimeout
```
odinserj: Why make it virtual? We can always change its value without any headache without this keyword.

rsilva93: This "mock":

https://github.com/rsilva93/Hangfire/blob/30373bf7117bc19f97827455ce103465d26fd228/tests/Hangfire.Core.Tests/States/ApplyStateContextFacts.cs#L30

Requires it to be virtual :)

src/Hangfire.Core/JobStorage.cs
rsilva93: Ok, I'll lock the get too.

src/Hangfire.Core/JobStorage.cs
odinserj: Hm, are there any other workarounds for tests? It's not a good reason to make a method virtual only to get tests working.

rsilva93: Maybe, just maybe, I can workaround the test I've write. But it become very common when mocking things for test purposes, because at the end, the mock is a kind of class inheritance. I can check my test later and try to not use this virtual if you do prefer :)

pieceofsummer:
I don't like the idea of letting expiration timeout to be changed during runtime. In my opinion, it should be a job storage option like `JobExpirationCheckInterval`, which is configured once during startup.

Here are my reasons:
- If the `JobExpirationTimeout` property is changed from multiple threads, you never know which value exactly it will have the moment the particular job changes its state, and guarding it with a lock won't change a thing. So it's just better not to change it at all.
- Even if it is only set during startup, it still requires explicit instantiation of a `JobStorage` subclass, which means we cannot use all those handy `.UseXyzStorage(connectionString, options)` extension methods.
- If we want different expiration times for different jobs, then we'll still need a job filter to make decisions at runtime.

rsilva93:
Hello @pieceofsummer.

Yeah, maybe it's not a good idea change this property multiple times during runtime. I was thinking about set this property only once at startup and then forget it.

About requires explicit instantiaton of the storage, that's not all true. We can always access job storage with the static Hangfire.JobStorage.Current, right?

And, it really doesn't exclude the filters funcionality. It's more about use a simple property of the storage to change some default config (that is hardcoded today) without need to add any filter if there is no need for different expiration times for jobs.

I think that this property can be at the options classes of the storage implementations in the future, so we can use it to set this value on startup and turn the property private to avoid the runtime changes, as you said.

Tell me what you think 😃

rsilva93:
@odinserj, I've turned the property non-virtual. Actually, I didn't need to mock it for tests :)

pieceofsummer:
> We can always access job storage with the static Hangfire.JobStorage.Current, right?

Somewhat true, but ugly. I'd actually be glad if all those `Current` static properties are eliminated for good because they're a thing of the past and don't fit into modern app design. And might be not true for NET Core at all, because `JobStorage.Current` is not necessary the one actually used.

Another good reason for this property to be located in storage options is because it's intuitive. Other properties related to job expiration and timeouts are already there, so it would be strange for `JobExpirationTimeout` not to.

rsilva93:
@pieceofsummer, you're totally right about the statics, but big part of Hangfire is still based on this. Hope we can improve it on 2.x (I think DI is the way 😄).

I agree with you about let this work for options classes, but they are specific for each storage and doesn't have a base class or interface where we can do it for now. We use Redis storage here, for example, and I think there is no way to improve it on his options without have the property somewhere on storage first (or maybe I just missed something).

odinserj:
Expiration timeout (hm, may be there's a better name for it?) is an important setting that is currently hidden in Hangfire internals, and the intention to move it to the configuration logic is quite good. I also agree with @pieceofsummer's position that it's not enough to add the corresponding property to the `JobStorage` class, because it is unclear how to use it without getting `JobStorage.Current` instance.

This setting does not relate to a specific storage, because it is declared and used in Hangfire.Core. We can't move it to `SqlServerStorageOptions` without creating a base class (`StorageOptions` or so), and I don't like to create classes.

I have two options to avoid using base class, and provide better configuration usability. Would be glad to hear your opinions.

1. Add `IGlobalConfiguration<JobStorage>.UseExpirationTimeout` extension method that will be used to configure the timeout.
    ```csharp
    GlobalConfiguration.Configuration
        .UseSqlServerStorage("HangfireConnection")
        .UseExpirationTimeout(TimeSpan.FromHours(4));
    ```
2. Add an overload to each `UseXXXStorage` method:
    ```csharp
    GlobalConfiguration.Configuration
        .UseSqlServerStorage("HangfireConnection", expirationTimeout: TimeSpan.FromHours(4));
    ```

pieceofsummer:
@odinserj adding overload for each UseXxxStorage is basically submitting a number of PR’s for all storage providers.

Don’t see how this would be any different from adding `JobExpirationTimeout` property into each provider’s `XxxStorageOptions` class (without need for them to have a common base class).

`IGlobalConfiguration<JobStorage>.UseExpirationTimeout()` extension method is a way to go, but still feels a bit counter-intuitive to me, since it would be the only option configured in a fluent way, while all the others are configured via XxxStorageOptions.

odinserj:
> Don’t see how this would be any different from adding JobExpirationTimeout property into each provider’s XxxStorageOptions class (without need for them to have a common base class).

@pieceofsummer, this PR works by setting the `ApplyStateContext.JobExpirationTimeout` property to `JobStorage.JobExpirationTimeout` in the `ApplyStateContext` class from the `Hangfire.Core` assembly.  How do you want to implement the idea you've mentioned above?

pieceofsummer:
@odinserj storage providers initialize the `JobExpirationTimeout` property in `XxxJobStorage`'s constructor from corresponding `XxxStorageOptions` property. Both this and your solution (adding an extra parameter to `UseXxxStorage` methods) would require them to link against a specific minimum version of Hangfire.Core, which adds `JobStorage.JobExpirationTimeout` property. 

This way, `JobStorage.JobExpirationTimeout` can be made read-only (with protected setter to initialize it from subclass' constructor), so there's no need to use locks for access synchronization.

dcorbett-fivebridgesllc:
It would be really nice to be able to override this timeout for a specific job type ( method call ), or even better within a job.   We have a class of jobs that can take hours to complete, while most jobs are much quicker.   A single value for both doesn't work well, and the solution right now seems to be creating separate servers to handle the appropriate job size.

rsilva93:
@dcorbett-fivebridgesllc, you can achieve it with filters:

https://discuss.hangfire.io/t/how-to-configure-the-retention-time-of-job/34

So you can have one or more filters setting different timespans and attribute them to your methods :)

pieceofsummer:
@dcorbett-fivebridgesllc this PR is about how long the **completed** jobs are preserved in the database, it has nothing to do with how long the job takes to finish. You've probably mistaken it with `InvisibilityTimeout`, which controls how long the job can be fetched before it is taken by another worker.

rsilva93:
@pieceofsummer and @odinserj, I like the XxxStorageOptions idea  (or even the ctor overload) because it's the prettier way, but not the easier to implement since we have multiple storages in separated repositories, and both ideas require to work on this multiple implementations.

I think we can do this:

1 - Add base class `StorageOptions` (can be useful in the future for other config's).

2 - Add 2 constructors in the `JobStorage`: one of them parameterless and another that accept a `StorageOptions` parameter. Or we can have only one constructor and force the storages implementation to pass an option, even with default values.

3 - Add a `StorageOptions` property in the `JobStorage`, with a protected setter. We will set the value in the constructor that receives the options parameter.

 4- In `ApplyStateContext` we can continue with the same idea, but reading the `ExpirationTimeout` from `JobStorage.Options.ExiprationTimeout` this time.

5 - We can already make `SqlServerStorageOptions` extends `StorageOptions` and the `SqlServerStorage`, that already accept a SqlServerStorageOptions, call the base JobStorage.ctor(options) (item 2).

After this, we can gradually update the other storages to use the same pattern. If you guys agree, I can do this implementation and send the PR's. I think it's a big change to solve a little problem, but gives Hangfire  some flexibility to new configurations like this (in a less painful way).

Tell me what do you guys think about it 😄

pieceofsummer:
We can actually combine the two ideas here:

1. Add a `JobExpirationTimeout` on the base `JobStorage` with a public getter and protected internal setter, and use its value in in `ApplyStateContext`.

2. Add an extension method `IGlobalConfiguration<JobStorage>.WithJobExpiration()` to temporary allow setting job expiration until all the job storages gradually add a property in theirs XxxStorageOptions. It will set timeout on `JobStorage.Current` for now.

3. Submit a number of PRs to corresponding storage repos to add an extra property in XxxStorageOptions and use it to initialize timeout in job storage constructors. After everyone has migrated to the new schema, the extension method above might be deprecated.

odinserj:
@pieceofsummer, regarding p.3, it will be required to bump `Hangfire.Core` dependency for each storage to the latest one. Customers will also be required to bump Hangfire.Core just to have consistent JobExpiration setting. This is a very high price. I'm proposing to implement only p.1 with public setter (some users don't use global configuration) and p.2.

@rsilva93, abstract class for options is a bit overkill. It's good to make abstraction if you have *not less than* two use cases. Unnecessary abstractions are pure evil.

This is a simple thing, let's keep it simple. May be a bit inconsistent, but simple.

pieceofsummer:
@odinserj regarding p.3, adding an extra parameter to `UseXxxStorage` overloads will also require bumping references to the latest Hangfire.Core, that's why I've said it's all the same :)

If we don't bump referenced Hangfire.Core version, the only solution is to provide an extension method for `IGlobalConfiguration<JobStorage>`. It would be a bit confusing, but it can be worked out later when the storages update to the latest version. 

And it should definitely become a storage option when migrating to Hangfire 2.0.

odinserj:
So, let's summarise the discussion, please add :+1: if you agree.

1. `JobStorage.JobExpirationTimeout` property should be added for those who don't use the global configuration class.
2. `IGlobalConfiguration<JobStorage>.UseJobExpirationTimeout(TimeSpan)` extension method should be added, and it should modify the corresponding property.
3. Regarding storage options – forget about them and ignore inconsistency until Hangfire 2.0.

rsilva93:
Ok, I got this 😃

rsilva93:
Hey guys, I think it's done.

I added the method `IGlobalConfiguration<TStorage>.WithJobExpirationTimeout`. Like @pieceofsummer 
had suggested, "With" sounds more fluent since it will be used after `UseXxxStorage` method: `UseStorage(x).WithJobExpirationTimeout(y)`.

odinserj:
Great @rsilva93! Now, the next thing is validation. I understand there was no any validation previously, but it's great chance to think about it in this PR. Values lower than zero are semantically wrong. But there may be problems with the values that are too low. Consider you set the expiration timeout to zero, and create a continuation in the following way:

```csharp
var id = BackgroundJob.Enqueue(/**/);
var continuation = BackgroundJob.ContinueWith(id, /**/);
```

With zero timeout, or worse, with timeout that equals to some seconds there's a chance that continuation will not be created, because antecedent job is already removed, and we can't query for its state. I know how to improve this in future version, but now I'm proposing to have a validation logic that throw an exception, when value is less than 1 hour to avoid such issues.

pieceofsummer:
Why use locks when getting/setting value for job expiration timeout? It is normally set only during initialization when no workers are running, so there won't be any cache coherency issues here.

rsilva93:
@odinserj, validation added at the propety setter, so it will cover the extension method too.

@pieceofsummer, it will be unusual, but the property will be public for change at runtime, so I think it's ok to have a lock there, just in case. 

Also: https://github.com/HangfireIO/Hangfire/pull/913#discussion_r120674167.

pieceofsummer:
@rsilva93 if the property is modified from multiple threads at runtime, it would be still unknown which value would be the current one at the moment `ApplyStateContext` is initialized, be it because of cache incoherency or just because of concurrency. 

Data caches will eventually become coherent, and I don't see any problem if *some* jobs will use an old value a bit longer. It's the same as running two server instances handling the same queue and changing expiration timeout for one of them. You have no idea which expiration timeout would be for the next processed job, so why care?

rsilva93:
Hey guys, sorry about the delay.

@pieceofsummer, I think you're right. We can't really ensure the consistency there even with the `lock`. I removed it.

rsilvanet:
Hey @odinserj, it will be considered for 1.7 release?

odinserj:
@rsilvanet, great! No worries regarding delays 😃 Yes, I'll merge these changes to the `dev` branch right now. Thank you!

PR #920, opened 2017-06-19
Permissions for dashboard actions

### LINKED ISSUE #758: Read only permission for the dashboard
As part of our deployment we will require that there is a read/view only role to the Dashboard.

similar to this forum request
https://discuss.hangfire.io/t/feature-request-read-only-dashboard/2514

mikechamberlain:
This is the pull request to close #758 and #423. It builds on and deprecates the current `Authorization` property by adding three configurable [`DashboardPermission`](https://github.com/agoda-com/Hangfire/blob/master/src/Hangfire.Core/Dashboard/DashboardPermission.cs)s:

- `DashboardPermission.ViewDashboard` - user can view the dashboard (equivalent to the existing `Authorization` property)
- `DashboardPermission.EnqueueJob` - user can trigger or enqueue jobs.
- `DashboardPermission.DeleteJob` - user can delete jobs.

Obviously, if `ViewDashboard` permission is not granted then the other two permissions are irrelevant.

## Configuration

To restrict permissions, we add filters to the new properties on the `DashboardOptions` dictionary like so:

````c#
var options = new DashboardOptions
{
    // deprecated: use ViewDashboardAuthorization property instead
    // Authorization = new[] { new ViewDashboardFilter() }
    ViewDashboardAuthorization = new[] { new ViewDashboardFilter() },
    DeleteJobAuthorization = new[] { new DeleteJobDashboardFilter() },
    EnqueueJobAuthorization = new[] { new EnqueueJobDashboardFilter() },
};
````

If one of the new permissions are not specified (`DeleteJob` or `EnqueueJob`) then for backwards compatibility it is granted by default.

## UI

Once a permission is denied, its associated elements are removed from UI. For instance, denying `DeleteJob` removes the all Delete Job buttons. Denying both these new permissions effectively results in a read-only view of the dashboard, so the selection checkboxes are also removed from the UI.

## AJAX calls

The associated permissions are also enforced on the Ajax calls, and the server will return a 422 if the required permission was denied.

## Unit tests

Instead of using dynamic Moqs (which seems to the preferred method in this codebase) to test the `DashboardContext`, I created stubs. As this class stands, with no virtuals or interface, it didn't seem possible to mock. I originally tried introducing an `IDashboardContext` interface, but this lead to breaking changes.

## Compatibility

I believe this version to be backwards compatible. I have tested it with the following extensions and packages:

- Hangfire.Autofac
- Hangfire.Pro
- Hangfire.Console

## Feedback please

Any feedback more than welcome. Would love to hear other people's ideas.

### CODE REVIEW

File: src/Hangfire.Core/Dashboard/DashboardOptions.cs [hunk 1]
```diff
@@ -16,17 +16,19 @@
 
 using System;
 using System.Collections.Generic;
-using Hangfire.Dashboard;
 
-namespace Hangfire
+namespace Hangfire.Dashboard
```
mikechamberlain: Saw no reason for the `DashboardOptions` class to not be in the Dashboard folder / namespace

mikechamberlain:
The latest commits address the problems noted above.

### Streamlined and unified authorizations

Combined "permissions" and authorizations into a single model, so we now declare all authorizations like this:

````c#
var options = new DashboardOptions
{
    // deprecated: use ViewDashboardAuthorization property instead
    // Authorization = new[] { new ViewDashboardFilter() }
    ViewDashboardAuthorization = new[] { new ViewDashboardFilter() },
    DeleteJobAuthorization = new[] { new DeleteJobDashboardFilter() },
    EnqueueJobAuthorization = new[] { new EnqueueJobDashboardFilter() },
};
````

### No breaking changes

I removed the `IDashboardContext` interface that I originally introduced for mockability.  Not only was this a breaking change for Hangfire Core, but I noticed it broke other extensions such as Hangfire.Console. This was no good. To work around this I created stubs instead.

### Documentation

I created a pull request for the docs repo outlining these new changes.

### What's next

I consider this ready to merge, unless anyone else would like to contribute?

mikechamberlain:
Closes #423.

odinserj:
@mikechamberlain, thanks a lot for your effort, especially for documenting the resulting changes. But the declared permissions aren't enough, because extensions may require another permissions. For example, [dashboard management](https://github.com/tracstarr/Hangfire.Core.Dashboard.Management) requires "schedule" and "add recurring job", batches from Pro have "Cancel" button, requiring the corresponding permission. So unfortunately I can't say this PR is complete.

To support fine-grained permissions there should be an extensible API. But (again!) in this case complexity of this feature is increased – we should be able to declare permissions, we should be able to register them somehow, we should be able to check for them.

I don't like complexity, and I'm trying to do best to avoid it where I can. Since I don't think this feature is very useful for 80% of people using this project, I can't evaluate the "fine-grained authorization rules" feature as good to be implemented. It's simple to implement a feature, but really hard to maintain it 😦 .

Have you considered to have only "Read Only" and "Full" authorization privileges? This feature is much simpler to implement and maintain. There's also a way to implement it without touching authorization filters at all, you can add the `Func<DashboardContext, bool> IsReadOnlyFunc` property to the `DashboardOptions` class.

mikechamberlain:
Thanks for the feedback @odinserj, I'll think about it and get back to you. Actually, `IsReadOnly` would be enough for our specific use-case too.

odinserj:
Great Mike, will wait then!

PR #933, opened 2017-07-07
Enable custom lock resources via DisableConcurrentExecutionAttribute

ian-a-anderson:
It is sometimes useful to be able to lock on an arbitrary resource that is shared across many different jobs. In our case we needed to lock access to the Dropbox API for single-write usage across several jobs (some recurring and some background). 

This adjustment to the DisableConcurrentExecutionAttribute facilitates shared locking by overriding the default resource name implementation with a customLockResource when specified. 

In order to accomplish our desired locking, we use this as follows:
`[DisableConcurrentExecution(20, "DropboxAPI")]`

### DIFF
--- a/src/Hangfire.Core/DisableConcurrentExecutionAttribute.cs
+++ b/src/Hangfire.Core/DisableConcurrentExecutionAttribute.cs
@@ -15,6 +15,7 @@
 // License along with Hangfire. If not, see <http://www.gnu.org/licenses/>.
 
 using System;
+using Hangfire.Annotations;
 using Hangfire.Common;
 using Hangfire.Server;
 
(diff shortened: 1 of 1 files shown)


odinserj:
I don't like the current `DisableConcurrentExecutionAttribute` because it blocks a worker from processing other jobs while waiting for a distributed lock to be acquired. I've written the [`MutexAttribute`](https://gist.github.com/odinserj/4a3bf40606c4da9183588a5a325dfb99) to use states instead. 

Another reason for using the latter attribute is because a distributed lock, acquired by the `DisableConcurrent...` filter can be released without any notification for the underlying code (due to connection loss, etc), making it less safe than should be.

ian-a-anderson:
Thanks for the update, I'll go ahead and close the Pull Request due to this guidance. 

Currently we run a specific server instance with a single queue and worker dedicated to processing these blocking jobs. This avoids any issues with locks blocking default job execution, at the expense of an increase in server polling chatter since we now have multiple "servers" per machine. It seems like even with MutexAttribute this approach would still be preferable to throwing these jobs into the default queue and letting them churn amidst other jobs.

If there's any way to optimize the polling process when multiple servers are running on a single machine, that would take away a pain point regarding our current implementation. Or, alternatively, allowing queues to have a per-queue worker thread limit would eliminate our need to run a separate server instance entirely. 

Hangfire is already amazing as-is today, so keep up the great work regardless!

odinserj:
@ian-a-anderson, you can use an overload of the `BackgroundJobServer` class' constructor to add a worker bounded only to one queue:

```csharp
app.UseHangfireServer(new Worker("dropbox"), new Worker("anotherapi"));
```

This will create a regular background processing servers with default number of workers listening the `default` queue, 1 worker listening the `dropbox` queue and 1 worker listening the `anotherapi` queue. Unfortunately those additional processes will not be shown in the dashboard, because I can't include this feature without introducing breaking changes to 1.X.

ian-a-anderson:
Awesome, good to know the functionality is in there already. Thanks!

PR #934, opened 2017-07-09
Read only view for dashboard

### LINKED ISSUE #758: Read only permission for the dashboard

mikechamberlain:
This is the pull request to close #758 and #423, and is a simplified version of my original PR #920. It adds the option to present a read-only view of the dashboard, preventing the user from deleting and enqueuing jobs, etc.

## Configuration

To enable read-only mode, we pass a function to the `DashboardOptions` like so:

````c#
var options = new DashboardOptions
{
    IsReadOnlyFunc = context =>
    {
        return true;
    }
};
````

## UI

Once read-only mode is activated, all delete and enqueue buttons and checkboxes are removed from the UI.

## AJAX calls

All AJAX calls that update state ensure that read-only mode is not activated.

## Unit tests

Instead of using dynamic Moqs (which seems to the preferred method in this codebase) to test the `DashboardContext`, I created stubs. As this class stands, with no virtuals or interface, it didn't seem possible to mock. I originally tried introducing an `IDashboardContext` interface, but this lead to breaking changes.

## Compatibility

I believe this version to be backwards compatible. I have tested it with the following extensions and packages:

- Hangfire.Autofac
- Hangfire.Pro
- Hangfire.Console

### CODE REVIEW

File: src/Hangfire.Core/Dashboard/DashboardOptions.cs [hunk 1]
```diff
@@ -16,16 +16,16 @@
 
 using System;
 using System.Collections.Generic;
-using Hangfire.Dashboard;
 
-namespace Hangfire
```
odinserj: It's better to rollback the namespace change, it's a breaking one

src/Hangfire.Core/Dashboard/DashboardRoutes.cs
```diff
@@ -95,8 +95,12 @@ static DashboardRoutes()
                 "/jobs/enqueued/fetched/(?<Queue>.+)",
                 x => new FetchedJobsPage(x.Groups["Queue"].Value));
 
-            Routes.AddClientBatchCommand("/jobs/enqueued/delete", (client, jobId) => client.ChangeState(jobId, CreateDeletedState()));
-            Routes.AddClientBatchCommand("/jobs/enqueued/requeue", (client, jobId) => client.ChangeState(jobId, CreateEnqueuedState()));
+            Routes.AddClientBatchCommand(
```
odinserj: Are there any changes in this file? If not, it's better to rollback the changes.

src/Hangfire.Core/Dashboard/Pages/AwaitingJobsPage.cshtml
```diff
@@ -48,31 +49,38 @@
...
-                            data-url="@Url.To("/jobs/awaiting/delete")"
-                            data-loading-text="@Strings.Common_Deleting"
-                            data-confirm="@Strings.Common_DeleteConfirm">
-                        <span class="glyphicon glyphicon-remove"></span>
-                        @Strings.Common_DeleteSelected
-                    </button>
-
+                    @if (!IsReadOnly)
+                    {
+                        <button class="js-jobs-list-command btn btn-sm btn-primary"
```
odinserj: I believe it's better to make buttons disabled with the corresponding reason in their tooltips instead of hiding them. This will explain to users what happened and how to enable those buttons. What do you think?

src/Hangfire.Core/Hangfire.Core.NetStandard.csproj
```diff
@@ -110,7 +110,7 @@
     <Compile Include="Common\TypeExtensions.cs" />
     <Compile Include="ContinuationsSupportAttribute.cs" />
     <Compile Include="Cron.cs" />
-    <Compile Include="DashboardOptions.cs" />
```
odinserj: Again, it's better not to move this class.

src/Hangfire.Core/Hangfire.Core.csproj
```diff
@@ -84,6 +84,7 @@
       <DependentUpon>Strings.resx</DependentUpon>
     </Compile>
     <Compile Include="Dashboard\DashboardContext.cs" />
+    <Compile Include="Dashboard\DashboardOptions.cs" />
```
odinserj: And again

samples/MvcSample/Startup.cs
```diff
@@ -1,4 +1,5 @@
-﻿using Hangfire;
+﻿using System.Collections.Generic;
```
odinserj: Could you rollback these changes also?

src/Hangfire.Core/Dashboard/CommandDispatcher.cs
```diff
@@ -39,6 +39,12 @@ public CommandDispatcher(Func<RequestDispatcherContext, bool> command)
 
         public Task Dispatch(DashboardContext context)
         {
+            if (context.IsReadOnly)
+            {
+                context.Response.StatusCode = 422;
```
odinserj: Hm, I believe "401 not authorized" response has a better reasoning what's happened, what do you think?

src/Hangfire.Core/Dashboard/Pages/AwaitingJobsPage.cshtml
mikechamberlain: I'm not sure really - I can see the argument for both choices.

I think it makes sense to disable a UI element when the user can do something to enable it. For example, in full access mode, the buttons disable until you check a checkbox. Here, because the user _can_ do something to make the action available, it makes sense to disable elements rather than hide them. These disabled elements give the user useful hints on what can be accomplished.

However, when in read-only mode, there is nothing the user can directly do to enable the action, 
 unless they are also the developer, in which case they will know why it is hidden and how to enable it.

If we _do_ choose to disable the buttons in read-only mode, I wonder what we do with the checkboxes. Should these be disabled too, each with a tooltip saying the same thing - "You don't have permission to perform this action - contact your system administrator for access"? It seems a lot of clutter. If we hide the checkboxes, but still disable the buttons, then this seems inconsistent.

Another point worth considering is that full-access mode is the default, meaning that if a user is given read-only privileges, then it's likely intentional. In this case, should we keep telling the user why they can't access a feature they probably should not have access to anyway?

For these reasons, in read-only mode I am leaning towards hiding elements rather than disabling. Of course, I will leave the final decision to you. Let you know what you would like me to do.

src/Hangfire.Core/Dashboard/DashboardOptions.cs
mikechamberlain: done

src/Hangfire.Core/Dashboard/DashboardRoutes.cs
mikechamberlain: done

src/Hangfire.Core/Hangfire.Core.NetStandard.csproj
mikechamberlain: done

src/Hangfire.Core/Hangfire.Core.csproj
mikechamberlain: done

samples/MvcSample/Startup.cs
mikechamberlain: done

src/Hangfire.Core/Dashboard/CommandDispatcher.cs
mikechamberlain: done

src/Hangfire.Core/Dashboard/Pages/AwaitingJobsPage.cshtml
odinserj: Good explanation, you persuaded me. I was also considering that with this option we can make dashboard accessible by default for all requests (currently only local ones are authorised), but make it readonly unless the request is a local one. 

Current authorisation defaults are safe, but sometimes confusing, when we deploy Hangfire to production and losing access to dashboard when forgot to configure authorisation rules. If we make it open, but readonly for all requests, initial setup will be much easier, and read-only privileges can lose the *intentional* point. That's why I've asked that question.

But looks like it's much better to leave authorisation as is, because dashboard even in the read-only access by default could be very harmful, because it opens potential security holes.

mikechamberlain: Agreed, it's better to be secure by default at the expense of convenience than view-versa.

PR #939, opened 2017-07-12
Add optional date bounds to recurring job

Rfvgyhn:
In reference to HangfireIO/Cronos/issues/3. Not sure if this is the _right_ way to do it, but this PR adds optional start and ends dates to recurring jobs. If specified, jobs are only queued if the current date is within the bounds of the specified dates.

### DIFF
--- a/README.md
+++ b/README.md
@@ -58,7 +58,7 @@ Usage
 
 This is an incomplete list of features; to see all of them, check the [official site](http://hangfire.io) and the [documentation](http://docs.hangfire.io).
 
-[**Fire-and-forget tasks**](http://docs.hangfire.io/en/latest/users-guide/background-methods/calling-methods-in-background.html)
+[**Fire-and-forget tasks**](http://docs.hangfire.io/en/latest/background-methods/calling-methods-in-background.html)
 
 Dedicated worker pool threads execute queued background jobs as soon as possible, shortening your request's processing time.
 
@@ -66,7 +66,7 @@ Dedicated worker pool threads execute queued background jobs as soon as possible
 BackgroundJob.Enqueue(() => Console.WriteLine("Simple!"));
 ```
 
-[**Delayed tasks**](http://docs.hangfire.io/en/latest/users-guide/background-methods/calling-methods-with-delay.html)
+[**Delayed tasks**](http://docs.hangfire.io/en/latest/background-methods/calling-methods-with-delay.html)
 
 Scheduled background jobs are executed only after a given amount of time.
 
@@ -74,7 +74,7 @@ Scheduled background jobs are executed only after a given amount of time.
 BackgroundJob.Schedule(() => Console.WriteLine("Reliable!"), TimeSpan.FromDays(7));
 ```
 
-[**Recurring tasks**](http://docs.hangfire.io/en/latest/users-guide/background-methods/performing-recurrent-tasks.html)
+[**Recurring tasks**](http://docs.hangfire.io/en/latest/background-methods/performing-recurrent-tasks.html)
 
 Recurring jobs have never been simpler; just call the following method to perform any kind of recurring task using the [CRON expressions](http://en.wikipedia.org/wiki/Cron#CRON_expression).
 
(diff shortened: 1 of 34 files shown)


odinserj:
@Rfvgyhn, please provide more details regarding this topic: what's added, how to use this and what's the impact on current users. I'd like to see this feature implemented, but need to ensure it's working fine.

burningice2866:
@Rfvgyhn you need to add explicit method overloads and not introduce new optional arguments since its effectively changing the signature and will cause MethodNotFound exceptions for existing code that hasn't been recompiled against this new method signatures.

aidmsu:
@Rfvgyhn Thank you for your PR! 

There are few questions:
1. Did you take into consideration that a user can pass `startTime`/`endTime` with different `DateTime.Kind` property: `Utc`, `Local` and `Unspecified`. What will be the behavior for each case?
2. Since we add `startTime`/`endTime` we should display them on the dashboard. Do you have any ideas how will it look like?

Rfvgyhn:
@odinserj 
* New overloads are added to `RecurringJob.AddOrUpdate` to allow for specifying a start and/or end date. Specifying these dates would queue new jobs only if the cron expression is matched and this run falls within the date bounds. 
* Specifying start and or end date creates an allowable window for jobs to be queued. Null start implies queue immediately like existing recurring jobs. Null end implies queue forever.
* I _think_ there's no impact on current users since the two fields are optional (after changing to explicit overloads like @burningice2866  mentioned). The job scheduler checks if start/end dates have a value. If not, the existing logic is used. 

@burningice2866 Will do.

@aidmsu 
1. An `ArgumentException` will be thrown if the start and end dates' kinds are different as specified by the [docs][0] since it [uses the same timezone][1] to convert to UTC. An `ArgumentException` will also be thrown if a non `UTC` kind is used without specifying the same in the `timeZone` param. I'm not sure what the ideal situation should be since I'm not too familiar with all the different DateTime edge cases. Should an exception be thrown stating the kinds need to be the same? Should it be enforced that they should already be UTC and it's up to the caller to make that conversion if needed? 
2. I hadn't given any though to the UI portion. I suppose two new columns could be added for start/end in the recurring jobs table. That would result in a lot of empty space for all jobs that don't specify start/end. There could also be one "bounds" column that has friendly text similar to the other date columns with a tooltip showing the exact dates. e.g. Assuming now = 2017-07-13:
    * Start: 2017-10-13, End: 2017-11-13, Text: In 3 months for 2 months
    * Start: 2017-01-01, End: 2017-11-13, Text: 2 more months 
    * Start: null, End: Null, Text: No bounds
Then there's the option of having the ability to see which recurring jobs exist but are beyond their end date so users could see which jobs are "expired" and won't ever be triggered again.

[0]: https://msdn.microsoft.com/en-us/library/bb382770(v=vs.110).aspx
[1]: /HangfireIO/Hangfire/pull/939/files#diff-97ecaffbd488a3bb085748fe1af5d108R77

Rfvgyhn:
Looking through the code a bit more, would it be a better approach to modify `ScheduleInstance` ctor to include a min and max date? Then `_schedule.GetNextOccurrences` could use the min and max dates which would ensure `NextInstant` and `GetNextInstants` returned the correct values.

Then the start/end date check could be moved out of the `RecurringJobScheduler` and the `_instantFactory` would be given start/end date instead.

yngndrw:
Would [DateTime.ToUniversalTime()](https://msdn.microsoft.com/en-us/library/system.datetime.touniversaltime(v=vs.110).aspx) be more suitable than [TimeZoneInfo.ConvertTime()](https://msdn.microsoft.com/en-us/library/bb382770(v=vs.110).aspx) ?

Rfvgyhn:
Another question. Adding explicit overloads for start and end dates results in 48 overloads for `AddOrUpdate`. Would it be better to create a new method (e.g. `AddOrUpdateBounded`) that uses optional params or is that many overloads OK?

ahanley:
Joe Coutcher has an implementation of this you should look at https://github.com/jcoutch/Hangfire.RecurringDateRange

pieceofsummer:
I’m closing this, since it is superseded by #952 now.

PR #944, opened 2017-07-18
Hangfire.SqlServer: check limits before inserting data

aidmsu:
Now you can get exception during invocation of `SqlServerWriteOnlyTransaction.Commit()` because one of methods `SetJobState`, `IncrementCounter` and so on tries to insert a string whose length exceeds the allowed value. It's hard to track the certain place of a bug.
Also you can get non-informative `SqlException` during invocation of `SqlServerConnection` methods such as `SetRangeInHash`, `SetJobParameter` due to the same reason.

This PR suggests to check if inserting string value exceeds allowable length:
1. If it's important field: state name, job parameter name, set key/value and so on then appropriate method throws `ArgumentException`.
2. If it's state reason field then the reason is truncated and a row is inserted to the table.

Also I feel we should have common limits for all storages in the future. It will be easier to change Hangfire storage without problems.

### DIFF
--- a/src/Hangfire.SqlServer/SqlServerWriteOnlyTransaction.cs
+++ b/src/Hangfire.SqlServer/SqlServerWriteOnlyTransaction.cs
@@ -101,7 +101,7 @@ values (@jobId, @name, @reason, @createdAt, @data);
 update [{_storage.SchemaName}].Job set StateId = SCOPE_IDENTITY(), StateName = @name where Id = @id;";
 
             var reason = state.Reason?.Length > Constants.StateReasonMaxLength 
-                ? TrancateReason(state.Reason) 
+                ? TruncateReason(state.Reason) 
                 : state.Reason;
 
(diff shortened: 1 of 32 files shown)


src/Hangfire.SqlServer/SqlServerWriteOnlyTransaction.cs
```diff
@@ -439,5 +480,10 @@ private void AcquireLock(string resource)
         {
             _lockedResources.Add($"{_storage.SchemaName}:{resource}:Lock");
         }
+
+        private string TrancateReason(string reason)
```
burningice2866: Minor spelling mistake; it should say `TruncateReason`

aidmsu: @burningice2866 oh, thanks)

alexsorokoletov:
@aidmsu @burningice2866 wondering if the conflict is the stopper for this to be merged?

Issue #871, opened 2017-04-20
KeyNotFoundException when accessing hangfire/jobs/deleted page


Gerras:
Version 1.6.8, Hangfire.SqlServer, 
Currently am seeing a KeyNotFoundException for certain Hangfire jobs when they have been deleted.

This is the stack trace:
``KeyNotFoundException 
[...]

PR #948, opened 2017-07-21
more complete fix for issue #871 - 

benjymous:
Don't crash when trying to access keys which don't exist (Issue #871)

Replaced the use of dictionary[key] with a minimally overridden version that returns default(TValue) instead of throwing - this fixes assorted failures in the query api when there are lots of old historical jobs with invalid function signatures, deleted data, etc.


odinserj:
Thanks @benjymous!

PR #952, opened 2017-07-26
Add optional date bounds to recurring job

Rfvgyhn:
This PR supersedes #939. I realized there's a dev branch that already has cronos merged. I went ahead and applied the same logic from the previous PR into the dev branch.

* New overloads are added to RecurringJob.AddOrUpdate to allow for specifying a start and/or end date in UTC. Specifying these dates would queue new jobs only if the cron expression is matched and this run falls within the date bounds (inclusive).
* Specifying start and or end date creates an allowable window for jobs to be queued. Null start implies queue immediately like existing recurring jobs. Null end implies queue forever.
* Start and end dates are stored in the recurring job dictionary which, in sql server, end up in the Hash table. I haven't tested with Redis but I assume the serializer automatically handles it just like the it did for sql server.
* I _think_ there's no impact on current users since the two fields are optional. The job scheduler checks if start/end dates have a value. If not, the existing logic is used.

Questions:
* Regarding @burningice2866's [comment][2], adding explicit overloads for start and end dates results in 48 overloads for AddOrUpdate. Would it be better to create a new method (e.g. AddOrUpdateBounded) that uses optional params or is that many overloads OK?
* Regarding @aidmsu's [comment][0] and my response (point 1), I have made it so the dates must already be in UTC. Is there a scenario where that wouldn't work?
* Does the UI portion need to be a part of this PR or can/should it be a part of another?
* What are the thoughts on my [previous comments][1](point 2) about UI?

[0]: https://github.com/HangfireIO/Hangfire/pull/939#issuecomment-315063423
[1]: https://github.com/HangfireIO/Hangfire/pull/939#issuecomment-315159243
[2]: https://github.com/HangfireIO/Hangfire/pull/939#issuecomment-315044477

### DIFF
--- a/src/Hangfire.Core/RecurringJob.cs
+++ b/src/Hangfire.Core/RecurringJob.cs
@@ -36,6 +36,27 @@ namespace Hangfire
             AddOrUpdate(methodCall, cronExpression(), timeZone, queue);
         }
 
+        public static void AddOrUpdate(
+            Expression<Action> methodCall,
+            Func<string> cronExpression,
+            DateTime startDate,
+            TimeZoneInfo timeZone = null,
+            string queue = EnqueuedState.DefaultQueue)
+        {
+            AddOrUpdate(methodCall, cronExpression(), startDate, timeZone, queue);
+        }
+
+        public static void AddOrUpdate(
+            Expression<Action> methodCall,
+            Func<string> cronExpression,
+            DateTime? startDate,
+            DateTime endDate,
+            TimeZoneInfo timeZone = null,
+            string queue = EnqueuedState.DefaultQueue)
+        {
+            AddOrUpdate(methodCall, cronExpression(), startDate, endDate, timeZone, queue);
+        }
+
         public static void AddOrUpdate<T>(
             Expression<Action<T>> methodCall,
             Func<string> cronExpression,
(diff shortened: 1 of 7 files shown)


ahanley:
I commented in a previous ticket but thought i'd comment here as well.  Joe Coutcher has an implementation of this you should look at https://github.com/jcoutch/Hangfire.RecurringDateRange

Rfvgyhn:
@odinserj @aidmsu @burningice2866 Any thoughts/feedback? I know the codecov check failed but the fix for that depends on keeping the 48 overloads or moving to a different method name with optional params.

burningice2866:
I personally like the idea of having a new `RecurringDateRangeJob` class used for scheduling recurring jobs with a start and end date. That way you don't need to pollute the existing `RecurringJob` class with tons of new overloads.

Rfvgyhn:
I moved the `RecurringJob` overloads to their own `BoundedRecurringJob` class. Any feedback is appreciated.

burningice2866:
I like it!

Rfvgyhn:
On a somewhat related note, the travis build failed and it looks like mono crashed right after running unit tests. Does that indicate a problem with the unit tests or an one-off issue with the build? Can the travis build be run again without having to push a new commit?

PR #956, opened 2017-07-31
Update Cron.cs

spookiecookie:
Added Cron.Daily TimeSpan support.

### DIFF
--- a/src/Hangfire.Core/Cron.cs
+++ b/src/Hangfire.Core/Cron.cs
@@ -78,6 +78,16 @@ namespace Hangfire
         }
 
         /// <summary>
+        /// Returns cron expression that fires every day at the specified timespan in UTC.
+        /// </summary>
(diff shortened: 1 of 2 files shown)


odinserj:
The `TimeSpan` class represents time intervals, not time of the day. And since there's an overload for `Cron` that allows you to specify hour and minute, it will be better to declare this method in your own project and use it without merging it into the main repository.

pieceofsummer:
@odinserj well, `DateTime.TimeOfDay` is a perfect example of `TimeSpan` representing time of the day :)
But you are right, it would be confusing if some method overloads will use argument as time interval while the others as exact time of the day.

burningice2866:
@pieceofsummer kinda works because the documentation states that it represents the time elapsed since midnight, aka a time interval. A better alternative to the TimeOfDay concept is to use `LocalTime` from the `NodaTime` library.

PR #974, opened 2017-08-20
Support async jobs returning ValueTask<T>

pieceofsummer:
C# 7 adds support for [generalized async return types](https://docs.microsoft.com/en-us/dotnet/csharp/whats-new/csharp-7#generalized-async-return-types), i.e. async method may not only return `Task` subclass but also an arbitrary object with `GetAwaiter()` method, e.g. `ValueTask<T>`.

### DIFF
--- a/src/Hangfire.Core/Server/CoreBackgroundJobPerformer.cs
+++ b/src/Hangfire.Core/Server/CoreBackgroundJobPerformer.cs
@@ -111,11 +111,16 @@ namespace Hangfire.Server
 
                 if (result != null && !methodInfo.ReturnType.GetTypeInfo().IsPrimitive)
                 {
+                    // Starting with C# 7, async methods can return any type that has an accessible GetAwaiter method. 
+                    // The object returned by the GetAwaiter method must implement the ICriticalNotifyCompletion interface.
+                    // Ref: https://docs.microsoft.com/en-us/dotnet/csharp/programming-guide/concepts/async/async-return-types
+
                     var getAwaiterMethod = methodInfo.ReturnType.GetRuntimeMethod("GetAwaiter", EmptyTypes);
-                    if (getAwaiterMethod != null)
+                    if (getAwaiterMethod != null && !getAwaiterMethod.IsStatic && getAwaiterMethod.IsPublic && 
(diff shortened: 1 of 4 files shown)


odinserj:
@pieceofsummer, the implementation based on `GetAwaiter` / `GetResult` methods is much better than the current one! However, we shouldn't invoke the `GetAwaiter` method before ensured its result type implements the `INotifyCompletion` interface.

pieceofsummer:
@odinserj I've added a more accurate validation of `GetAwaiter()`. 

I'm a bit confused about `ICriticalNotifyCompletion` though because [some sources](https://blogs.msdn.microsoft.com/pfxteam/2011/01/13/await-anything/) say it is optional, while [the others](https://docs.microsoft.com/en-us/dotnet/csharp/programming-guide/concepts/async/async-return-types) say it is mandatory. The latter article is more recent, so I'd better comply with it.

odinserj:
Indeed, interesting question. All of these interfaces were added in .NET Framework 4.5, and I [don't see](https://source.dot.net/#System.Private.CoreLib/shared/System/Runtime/CompilerServices/INotifyCompletion.cs,0670f9b4b67cecd6,references) any direct `INotifyCompletion` interface implementations in .NET Core, but unfortunately can't check this on [reference source](https://referencesource.microsoft.com/#mscorlib/system/runtime/compilerservices/TaskAwaiter.cs,be57b6bc41e5c7e4) to learn about .NET Framework (it isn't browsable there). 

I think we should either ensure that `INotifyCompletion` interface is not enough to an awaiter object awaitable, or do use it, because it's more general.

Another point – we shouldn't invoke the `GetAwaiter` method if there's no `GetResult` method. I.e. don't do anything, if a method doesn't satisfy all the requirements to be awaitable.

pieceofsummer:
Enforced a full set of checks on `GetAwaiter` return type:
- Implements `INotifyCompletion` interface
- Has a boolean `IsCompleted` instance property with a public getter
- Has a public parameterless `GetResult` instance method

odinserj:
... and this definitely great PR has finally merged 😃

PR #976, opened 2017-08-21
Cancel the CancellationToken when the job is deleted from dashboard

pieceofsummer:
So running async tasks can gracefully abort


plaisted:
I've often started by job methods with something like:
```csharp
var source = new CancellationTokenSource();
Task.Run(() => JobCancellationHelper(source, jobCancellationToken, 5000)); //jobCancellationToken is hangfires IJobCancellationToken  
```
and then pass the new `CancellationToken` from the `CancellationTokenSource` instead of using the Hangfires `IJobCancellationToken`.  The `JobCancellationHelper` is a simple method like:

```csharp
public async Task JobCancellationHelper(CancellationTokenSource source, IJobCancellationToken cancellationToken, int delayMs)
{
    while (!source.Token.IsCancellationRequested)
    {
        try
        {
            //requires DB round trip
            cancellationToken.ThrowIfCancellationRequested();
        }catch (OperationCanceledException e)
        {
            source.Cancel();
            return;
        }
        await Task.Delay(delayMs, source.Token).ContinueWith(t=> { });//swallow exception
    }
}
```

Having Hangfire handle that for me would be nice. Haven't looked at your suggested changes but just posting this here in case others have this issue and want a workaround.

pieceofsummer:
@plaisted I believe there's a serious issue with your snippet because multiple threads are using the same `IStorageConnection`, which could lead to race conditions and internal state corruption. 

I also don't see how the task terminates when the job finishes normally. Do you need to manually call `source.Cancel()` when the job finishes?

plaisted:
@pieceofsummer 
Thanks for the comments.  Regarding calling `source.Cancel()` I've edited the code snippet I put there to reflect cancelling and disposing of the CTS.

As far as the shared `IStorageConnection` I'll look into the code for the `Worker` a little more.  My assumption was that if it is safe to call in the main method it should be safe to call in a spawned task as long as you don't access it in both. I think the `JobContext` and `IJobCancellationToken` share a single `IStorageConnection` so there may be conflicts there if using both. Can you elaborate on the issue a little more? Looking at SqlServer it appears to create a new Sql Connection for each command so not sure there would be an issue there.

pieceofsummer:
@plaisted sure, it may only affect the situation when you use it from the job code too, e.g. logging to Hangfire.Console etc. 

SQL Server storage really creates a new `SqlConnection` every time, but there's no guarantee other storage providers will do the same.

Also, have to point out that `IJobCancellationToken.ThrowIfCancellationRequested()` may throw exception other than `OperationCanceledException` (for example, if some database error occurred), so you probably need to handle those too.

rossknudsen:
How about rewriting @plaisted's solution with a class implementing `IDisposable`?  But it would be nicer if Hangfire could replace `IJobCancellationToken` with a `CancellationToken`.  Any comments on the following would be appreciated though.

```
public class JobCancellationTokenWrapper : IDisposable
{
    private readonly CancellationTokenSource _cancellationSource = new CancellationTokenSource();

    public JobCancellationTokenWrapper(IJobCancellationToken jobCancellationToken, int delayMs = 5000)
    {
        Task.Run(async () =>
        {
            while (!_cancellationSource.Token.IsCancellationRequested)
            {
                try
                {
                    jobCancellationToken.ThrowIfCancellationRequested();
                }
                catch (OperationCanceledException e)
                {
                    _cancellationSource.Cancel();
                    return;
                }

                await Task.Delay(delayMs, _cancellationSource.Token).ContinueWith(t => { }); //swallow exception
            }
        });
    }

    public CancellationToken Token => _cancellationSource.Token;

    public void Dispose()
    {
        _cancellationSource.Cancel();
        _cancellationSource.Dispose();
    }
}
```
Usage:
```
public async Task Execute(int snapshotId, IJobCancellationToken cancellationToken)
{
    using (var wrapper = new JobCancellationTokenWrapper(cancellationToken, 5000))
    {
        // TODO implement work here.
        // pass from wrapper.Token to methods that accept CancellationTokens
    }
}
```

odinserj:
@pieceofsummer thanks for this awesome pull request :heart: Finally merging!

PR #993, opened 2017-09-13
Add JobDisplayNameAttribute

pieceofsummer:
This should have been done a long time ago :)

Supports formatting:
```c#
[JobDisplayName("Test job ({0})")]
public static void TestJob(int arg)
```

Supports localization:
```c#
[JobDisplayName("TestJob_DisplayName", ResourceType = typeof(MyResources))]
public static void TestJob(int arg)
```


odinserj:
Finally merged, @pieceofsummer, thanks for your efforts! 👍 👍 👍

pieceofsummer:
@odinserj I wonder if this should take precedence before display name provider? If yes, then NETFULL `DisplayNameAttribute` should too.

Looking back at this PR, I also think maybe we should unseal `JobDisplayNameAttribute` and make `Format` method virtual, so users may easily inherit custom attributes from it and implement their own naming. What do you think?

odinserj:
Right! I've just made the corresponding commits to implement these ideas.

pieceofsummer:
Great! Now I think the only thing left (before the major update to fully async model in 2.0) is to cancel the `CancellationToken` when the job is deleted from dashboard :)

odinserj:
I'll mark it with the 1.7.0 milestone then and check a bit later ("later" sounds dangerous, he-he) 😃

ernecastro:
Hi all, I don't if I'm missing something, but this feature does not seem to work for me.

Having this method:

`        [JobDisplayName("Test job ({0})")]
`
`       public static void ExecuteJob(int p)
`
`       {
``          ...
`       }

 
Gives this name in the dashboard:

  Test job (System.Object[])

pieceofsummer:
@ernecastro see #1136. Fixed in `dev` branch already, will be included in the next beta.

rfrancisco:
When can we expect the next beta with this fix?

joshuagriffin:
We are waiting for this as well

Shamshiel:
Does this only work with static methods? I'm not able to get it to work with my non-static methods.

freeflyrw:
Same issue for me. @Shamshiel. @pieceofsummer No joy. Method call below not static. Had you any luck with this?

[JobDisplayName("Perform CatNet Analysis - CatNetJobQueueId {0}")]
Task PerformCatNetAnalysis(string catNetJobQueueId, string userEmail, string locationName);

freeflyrw:
Ahhh...had the attribute at interface rather than class level. All good now!

taylor-viventium:
This seems to be broken. It has no effect on my job name. Tried both DisplayName and JobDisplayName with a string value.

odinserj:
If you are using a base class or interface for your background jobs, please try putting the attributes there instead of an implementation class.

Issue #828, opened 2017-03-08
JobActivator BeginScope with the full PerformContext


TheDadycool:
Hi,

It will be nice to have in the JobActivator a BeginScope function taking the full PerformContext instead of the JobActivatorContext... This will allow to register the PerformContext in the scope right away, and using it later in the actual job function. The backward compatibility will be really easy by just calling the previous one by generating the JobActivatorContext like in: 

public virtual JobActivatorScope BeginScope(PerformContext context)
{
return this.BeginScope(new JobActivatorContext(context.Connection, context.BackgroundJob, context.CancellationToken)));
}

Regards,

David.

KLuuKer:
I would also like to have this change because this would allow me to remove any direct dependency to hangfire in my jobs project.

Here is my example:
- MyBigApp.Jobs: containing a interface called IMyJobContext and only the actual job code, with the jobs constructor that accepts the IMyJobContext
- MyBigApp.Jobs.HangFireService: this runs the hangfire server and dashboard, and supplies a implementation of IMyJobContext that uses PerformContext internally to allow it to track various state information (and for example the Hangfire.Console package functionality)
- MyBigApp.Backend: this is the backend of my big application and is the one that is responsible for adding the Jobs into the database using hangfire

this way the backend project only needs a reference to Hangfire.Core and Hangfire.SqlServer
without the extra reference caused (for example) by the Hangfire.Console package

And if Hangfire in the future would get better separation from itself so that only the bit of code that actually pushes jobs into the database needs to be referenced, my backend project would have even less unnecessary references.

And when you have a mixed frameworks, like HangFireService running on net461 and the Backend running on .net core, it becomes a even bigger pain to have all these dependencies everywhere.
Or if a extension to Hangfire like Hangfire.Console not being converted (or very just slowly) to .net core causing the conversion of the Backend project to run on .net core being blocked.

jeroenvervaeke:
As mentioned above, this would be a very great change.

Here is my example:
- MyBigApp.Backend -> Contains a datalayer, business layer, service layer + everything is resolved using windsor, logging is done using NLog using the ILogger interface.
- MyBigApp.Hangfire -> Contains the jobs, Hangfire.Windsor resolves dependencies (businesslayer from MyBigApp.Backend)

In my example both the Backend and Hangfire project call the business layer.

With this change I would be able to forward all the calls made to ILogger (made from within MyBigApp.Hangfire) to the Hangfire dashboard.

[...]

PR #995, opened 2017-09-14
Fixing #828

jeroenvervaeke:
Provides the JobActivator BeginScope with the full Performcontext.
Examples and motivation can be found here: https://github.com/HangfireIO/Hangfire/issues/828

This change is fully backward compatible with previous versions

### DIFF
--- a/src/Hangfire.Core/JobActivator.cs
+++ b/src/Hangfire.Core/JobActivator.cs
@@ -17,6 +17,7 @@
 using System;
 using System.Collections.Generic;
 using Hangfire.Annotations;
+using Hangfire.Server;
 
 namespace Hangfire
 {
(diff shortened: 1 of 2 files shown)


odinserj:
This is a nice change, and I was wrong with adding the new context class just for IoC, because `PerformContext` allows much better usage, and we could also register it in an IoC container scope. Thank you for contribution, this change will be available in 1.7 (I will merge it with the `dev` branch).

KLuuKer:
@odinserj would it be possible to get this backported to 1.6.x?
I'm currently really breaking my neck jumping around\trough hoops (on fire) to get a half decent log abstraction going (appinsights\hangfire console\ilogger).

KorsG:
I would also very much like to see this implemented in 1.6.x, since it seems 1.7.0 is a big release and alot of work is still required before it is released.

PR #1003, opened 2017-09-20
Transaction management without TransactionScope

aliashab:
Transaction management without TransactionScope

https://discuss.hangfire.io/t/transaction-management-without-transactionscope/1128

use

```csharp
var mySqlConnection = new SqlConnection(connectionString); 
mySqlConnection.open();
var tran = mySqlConnection .BeginTransaction();
var storage = new SqlServerStorage(mySqlConnection, tran); 
var client = new BackgroundJobClient(storage);
client.Enqueue(() => Run(queueEntity.RecId));
tran.Commit();
```


Singularity222:
Any idea when this pull request might be completed and merged into a new version? This feature would help me out a lot.

Singularity222:
Is there any chance this pull request could be merged and a new version of Hangfire sent to Nuget containing these changes?

odinserj:
@aliashab, thanks for your contribution! I think it's better to modify the `UseConnection` method to provide both `DbConnection` and `DbTransaction` parameters for its action, because otherwise things are messed a bit: connection is taken from a parameter, transaction is from a global scope.

```cs
_storage.UseConnection((connection, transaction) =>
{
    connection.Execute(/* ... */);
});
```

In this case the `UseTransaction` method will always create a new transaction, and the `UseConnection` method will allow to use an existing one.

odinserj:
@Singularity222, there will be no quick fix for this, the change is planned to the next version only. To speed up the things, you could build your own version from sources, I understand that this is more complex usage.

Ppiwow:
Any news on that one?

ulrichb:
Oh, yeah, would exactly need this for integrating Hangfire into a legacy application which uses `DbTransaction`s instead of `TransactionScope`. Please consider this for 1.8.

ulrichb:
@odinserj As this PR is outdated and has some conflicts: If you lik,e I could pick it up, rebase it on the current master and fix the build.

ddiosdado-tea:
Any updates on this? it would be great to treat internal work living in the same db and hangfire operations as a single unit

PR #1006, opened 2017-09-23
Fix number of recurring jobs per page

pieceofsummer:


odinserj:
Thanks Alex! There was PR #832 that "fixed" the opposite issue, but looks like the problem itself was related not to Dashboard, but rather to a storage.

PR #1018, opened 2017-10-06
Update rickshaw css to match js

pieceofsummer:
Seems like css was not updated along with js, so tooltip position adjustment near the graph's right edge never worked, which caused the horizontal scroll bar to appear.


odinserj:
Great, thanks for polishing this!

PR #1029, opened 2017-10-20
Fix server start time tooltip position

pieceofsummer:
Before:
screen shot 2017-10-20 at 20 44 18

After:
screen shot 2017-10-20 at 20 44 56


odinserj:
Thanks @pieceofsummer, looks much better! There's one unit test that's failing, but it is related to SQL Server. I've added a diagnostic message to identify the problem in future.

PR #830, opened 2017-03-09
[WIP] Implemented Manual Job

reaction1989:
see #382

Rewritten the commits of @gandarez see 
https://github.com/gandarez/Hangfire/tree/manual-job

Changes are not tested by now.

Issue #382, opened 2015-05-28
"Manual launch" jobs or recurring jobs without schedule


geirsagberg:
It would be really cool if one could register jobs that will show up in the dashboard but can only be launched manually, e.g. `ManualJob.AddOrUpdate<IMyService>(service => service.DoOneTimeStuff())`.

I have tried hacking it with e.g. recurring jobs with CRON expressions for 31st of February, but Hangfire crashes.

smhinsey:
I'd be interested in this as well.

vip32:
interesting how you want to pass in the job arguments then. the dashboard should also facilitate that.
the feature sure is interesting, in my current project we build a mvc pages which can handle our non scheduled job starting. another scenerio is making a POST request to the dashboard to start a job, so a more JS style app can take care.

kataik:
I'd like to have something like that as well, for example
`jobId = BackgroundJob.Stash<IMyService>(service => service.DoOneTimeStuff());`
`DoOtherStuff();`
`BackgroundJob.Enqueue(jobId);`

gandarez:
@kataik what did you mean with "stash"? Does it have the same meaning of "Manual Update"?

gandarez:
+1

kataik:
Hi Carlos,
I mean it would be nice to have something scheduled somewhere, but
trigerred manually - even from a different service. If you mean something
like that on "Manual Update", then yes. :)

Best,
Kornél

2015-11-29 23:05 GMT+01:00 Carlos Henrique Guardão Gandarez <
notifications@github.com>:

> +1
>
> —
> Reply to this email directly or view it on GitHub
> <https://github.com/HangfireIO/Hangfire/issues/382#issuecomment-160475354>
> .
>



-- 
Kornél Bálint Kátai
Software Developer

Sonrisa Kft
H-1022 Budapest, Bég u. 3-5., HUNGARY

gandarez:
@kataik I'll try to do something related to it into my branch "manual-job"
https://github.com/gandarez/Hangfire/tree/manual-job

gandarez:
Hey guys I've done 99% of the new Manual State job. Please refer to https://github.com/gandarez/Hangfire/commit/830f3ebe87ad2432959978bdca7ab4cfe201b8a7

vip32:
a meaningful addition. will your solution be merged? seems like it cannot work as a standalone extension.

gandarez:
[...]

PR #1036, opened 2017-11-01
[WIP] Implemented Manual Job (With tests)

BeeHiveJava:
See #830. All credits go out to @reaction1989, who provided the original implementation. All I did is added tests in hope to get this feature going sooner!

> see #382
> 
> Rewritten the commits of @gandarez see
> https://github.com/gandarez/Hangfire/tree/manual-job
> 
> Changes are not tested by now.


BeeHiveJava:
@odinserj This line specifically fails for some reason, it happened in another PR too (#1029), not too certain why.
https://github.com/HangfireIO/Hangfire/blob/6525aa314f65d1a029863928edf8c9cfc18ec30d/tests/Hangfire.SqlServer.Tests/SqlServerWriteOnlyTransactionFacts.cs#L55

danielcor:
I don't get it.   What about this is so different than already exists?

BeeHiveJava:
@danielcor, As @gandarez once explained: a manual job has to be manually triggered from the dashboard (or code) in order to be executed. Unlike a recurring or scheduled job, it does not execute on a predefined time.

In our usecase this is very useful for testing and cleaning purposes.

dcorbett-fivebridgesllc:
Normal jobs are triggered from code to be executed….  And this extension allows one to manually queue jobs:

https://github.com/tracstarr/Hangfire.Core.Dashboard.Management

This enhancement doesn’t require a modification to Hangfire.
So I still don’t understand the benefit.

BeeHiveJava:
I can't believe we've overlooked that one. That seems to be about perfect for our use case, thanks! I'll make sure to comment it on the other PR as well, so that future searchers may be helped out perhaps.

PR #1056, opened 2017-11-10
Added an extension method that returns a normalized method name

### LINKED ISSUE #1050: Job.FromExpression fails to resolve method for explicit interface implementations
```c#
[Fact]
public void TestExplicitInterface()
{
    IService service = new ServiceImpl();
    var job = Job.FromExpression(() => service.Method()); // throws
}

public interface IService
{
    void Method();
}

public class ServiceImpl : IService
{
    void IService.Method()
    {
    }
}
```

f00:
fixes #1050


pieceofsummer:
This is an incorrect implementation, because a method with the same name and parameters could be implemented both explicilty and implicitly, and you’re risking to select a wrong one.

The most common example of this is `GetEnumerator()` for generic collections, which is typically implemented explicitly for `IEnumerable` and implicitly for `IEnumerable<T>`.

The correct implementation would be to use InterfaceMap if the method is called on interface-typed variable.

odinserj:
@pieceofsummer, this is the only solution that makes sense and is consistent with the current behavior. Currently, when the non-generic `FromExpression` method is called with a variable from the current closure, the type of the newly created job will point to the concrete type, not an interface (`ServiceImpl`, not `IService` from the referenced issue) due to the historical reasons. So the resulting job will look like:

```
{
    "type": "MyNamespace.ServiceImpl, MyAssembly",
    "method": "Method"
}
```

Since the concrete type is used, any explicit interface implementation methods will be non-public. And since we don't support other non-public methods (with `private`, `internal`, etc. modifiers), the only solution will be to disallow this behavior, and throw the `NotSupportedException` like in other cases with non-public methods.

odinserj:
Marcus (@f00), I've also added a unit test in commit b9023f31a5d852eb2ef73c1f439681316a6fdcac to check the resulting behavior, when explicit interface implementation is used. So this PR is finally merged, thank you for your work! ❤️

Issue #1078, opened 2017-12-17
Dashboard footer should not display the connection string or it should be optional


sandorfr:
Currently, the connection string is displayed in the footer. It shouldn't or at least it must be very easy to disable this behavior. 

PR : #1079

odinserj:
Hangfire dashboard is designed to be an internal part of your application that is accessible only for developers/devops staff. Could you tell me why they shouldn't see the connection string?

sandorfr:
In my case only the accredited people have access to those information. People with access to the administration backend don't have access to the database or the redis or even the filesystem. They don't need it and the should not have it. People operating the software are not people maintaining the infrastructure. I can give you a simple scenario. Someone in the team operating the software quits, if this kind of information has been exposed, it needs to be changed in addition to his account revoked.

PR #1079, opened 2017-12-17
add on option to remove the storage connection string from the footer…

sandorfr:
… #1078

### DIFF
--- a/src/Hangfire.Core/Dashboard/RazorPage.cs
+++ b/src/Hangfire.Core/Dashboard/RazorPage.cs
@@ -41,8 +41,7 @@ namespace Hangfire.Dashboard
 
         public JobStorage Storage { get; internal set; }
         public string AppPath { get; internal set; }
-        public int StatsPollingInterval { get; internal set; }
-        public bool DisplayStorageConnectionString { get; private set; }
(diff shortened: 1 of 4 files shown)


src/Hangfire.Core/Dashboard/RazorPage.cs
```diff
@@ -93,6 +94,7 @@ internal void Assign(DashboardContext context)
             Storage = context.Storage;
             AppPath = context.Options.AppPath;
             StatsPollingInterval = context.Options.StatsPollingInterval;
+            DisplayStorageConnectionString = context.Options.DisplayStorageConnectionString;
```
pieceofsummer: Should be initialized in another `Assign()` overload too.

sandorfr: Done ;)

pieceofsummer:
I also think it's no longer reasonable to add an extra page property for each Options property of interest (there's already two of them, and even more after this PR).

Would make sense to add the full `Options` property instead, and dereference its subproperties in the page code.

odinserj:
Thanks guys for PR and for reviewing it. I'm merging these changes, they will be released with 1.6.18 next week, since they are related to security.

PR #1084, opened 2017-12-24
Deprecate `Cron.XXXInterval` methods

### LINKED ISSUE #1041: Deprecate `Cron.XXXInterval` methods
Neither of these methods works correctly, when the specified interval is not a multiple of the corresponding field's range length (max-min). For example, when we are using `Cron.HourInterval(10)`, we will not get the documented *every 10 hours* semantic, instead we'll get the following occurrences:

```
Nov 6, 00:00
Nov 6, 10:00
Nov 6, 20:00 
Nov 7, 00:00
```

Time interval between the last two entries is only 4 hours and violates what's written in the method's name and in its comments. Cron expressions don't allow to specify "every N" recurring intervals in general, and other methods should be used instead of confusing methods. These methods should be deprecated with the `ObsoleteAttribute` and removed in the next major version.

almez:
Deprecate the following methods:

string MinuteInterval(int interval);
string HourInterval(int interval);
string DayInterval(int interval);
string MonthInterval(int interval);

Fixes #1041.

### DIFF
--- a/src/Hangfire.Core/Cron.cs
+++ b/src/Hangfire.Core/Cron.cs
@@ -218,7 +218,7 @@ namespace Hangfire
         /// Returns cron expression that fires every &lt;<paramref name="interval"></paramref>&gt; minutes.
         /// </summary>
         /// <param name="interval">The number of minutes to wait between every activation.</param>
-        [Obsolete("Please use `ScheduledState` instead. Will be removed in 2.0.0.")]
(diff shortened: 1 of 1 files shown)


odinserj:
Congrats with your first contribution! I've changed the to the `dev` to push these changes to 1.7.0 version only, and will leave some review comments in a while.

almez:
sure, it has been updated.

odinserj:
@almez, thanks for handling this, merged! 👍

PR #1087, opened 2018-01-09
Update README.md to mention Razor Generator

sandorfr:
Mention that Razor Generator is required to modify the Razor views.


odinserj:
Great, thanks!

PR #1093, opened 2018-01-17
Fixing issue 1092 - arguments values are duplicated in SQL Server Job table

dezfowler:
- Adding failing test which highlights the problem - arguments are duplicated
- Fixing by copying the reference from the InvocationData object's Arguments property and nulling it out so it's not serialized


odinserj:
Hi @dezfowler, this issue is fixed in the recent [1.7.0-beta1](https://github.com/HangfireIO/Hangfire/releases/tag/v1.7.0-beta1) release. I left the arguments in the job payload itself, to have the same payload structure also for recurring jobs. For backward compatibility, that column is still used when reading information for old background jobs.

PR #1114, opened 2018-02-19
Add AddHangfire method overload to allow resolve connection string fr…

denis-ivanov:
…om another dependency in container

For now, we must strictly specify connection string when configure container.
In my pull request, we can get connection string from dependency in container like this:

```csharp
services.AddHangfire((provider, config) =>
{
    var settings = provider.GetService<ISettings>();
    var connectionString = settings.ConnectionString;
    config.UseSqlServerStorage(connectionString);
});
```


denis-ivanov:
The same method have Entity Framework Core - https://github.com/aspnet/EntityFrameworkCore/blob/093e30fdd83776bda0275c27cf596f390c3ac763/src/EFCore/EntityFrameworkServiceCollectionExtensions.cs#L222

odinserj:
Thanks for the PR, @denis-ivanov! I've added this feature as a part of 1.7.0-beta3 and included a credit to you in the [release notes](https://github.com/HangfireIO/Hangfire/releases/tag/v1.7.0-beta3). When I'll backport this feature to 1.6.X, I will add you as a co-author to the commit message.

Issue #796, opened 2017-01-19
Hangfire 0 active servers after SQL Server outage, won't start servers again


tompazourek:
Yesterday, my SQL Server was offline for about 40 minutes.

I have 6 instances of Hangfire 1.6.4 running on various websites. When my SQL Server was down, Hangfire was throwing a lot of exceptions (which is expected), but after it came back, 5 out of 6 of those instances reported 0 active servers. I could go to the dashboard and go through everything, but even when the SQL Server was up again, the Hangfire servers wouldn't start.

Also, the exceptions from Hangfire were no longer being thrown. There were just no active servers.

I waited for like 20 minutes, and the Hangfire servers still didn't start, so I had to manually restart each of those 5 websites.

The exceptions thrown by Hangfire were about the SQL Server connectivity:

```
System.Data.SqlClient.SqlException (0x80131904): A network-related or instance-specific error occurred while establishing a connection to SQL Server. The server was not found or was not accessible. Verify that the instance name is correct and that SQL Server is configured to allow remote connections. (provider: Named Pipes Provider, error: 40 - Could not open a connection to SQL Server) ---> System.ComponentModel.Win32Exception (0x80004005): The network path was not found
   at System.Data.SqlClient.SqlInternalConnectionTds..ctor(DbConnectionPoolIdentity identity, SqlConnectionString connectionOptions, SqlCredential credential, Object providerInfo, String newPassword, SecureString newSecurePassword, Boolean redirectedUserInstance, SqlConnectionString userConnectionOptions, SessionData reconnectSessionData, DbConnectionPool pool, String accessToken, Boolean applyTransientFaultHandling)
   at System.Data.SqlClient.SqlConnectionFactory.CreateConnection(DbConnectionOptions options, DbConnectionPoolKey poolKey, Object poolGroupProviderInfo, DbConnectionPool pool, DbConnection owningConnection, DbConnectionOptions userOptions)
   at System.Data.ProviderBase.DbConnectionFactory.CreatePooledConnection(DbConnectionPool pool, DbConnection owningObject, DbConnectionOptions options, DbConnectionPoolKey poolKey, DbConnectionOptions userOptions)
   at System.Data.ProviderBase.DbConnectionPool.CreateObject(DbConnection owningObject, DbConnectionOptions userOptions, DbConnectionInternal oldConnection)
[...]

PR #1117, opened 2018-02-23
Added server announcement to Heartbeat if the server is found not present

SaDa1337:
#796 
Added server announcement to Heartbeat if the server is found not present 

Description:
Whenever the server tries to update the heartbeat with the storage it may have been previously deleted by the timeout watcher - because of a sql timeout (server down during the watchdog trigger). 

Implementation:
I've added a check if the server is present in the storage during heartbeat check, and if it's missing will announce it again.


odinserj:
Thanks for you work on this (and don't look at a lot of commits). This problem is solved by PR https://github.com/HangfireIO/Hangfire/pull/1082, released with [1.7.0-beta1](https://github.com/HangfireIO/Hangfire/releases/tag/v1.7.0-beta1) by using a two-step process:

1. If a server is unable to send heartbeats for the time specified in ServerTimeout option, it will give up and attempt to restart itself with the newly generated id, because it's likely that it will be already removed by other servers when connection is re-established.
2. If the `Heartbeat` command caused the `ServerGoneException`, which is now thrown when a server was already marked as dead by other servers, the server will be restarted with the newly generated id.

PR #1121, opened 2018-02-27
Resolves #635 by adding a custom "DashboardInfo" attribute.

### LINKED ISSUE #635: DisplayName not supported in .NET Core?
I don't understand why DisplayName logic is excluded from .NET Core builds in HtmlHelper (using `#if NETFULL`). It seems to me like it should be possible to include that code; `DisplayNameAttribute` and reflection machinery are available in .NET Core.

I'm building an ASP.NET Core 1.0 app with Hangfire and would love having this. The dashboard is much less useful without that feature.

josiahbrower:
The "DashboardInfo" attribute will work for .NET Core projects, and can be used in .NET Framework projects, but is backwards compatible because the "DisplayName" attribute will still work in .NET Framework projects.


pieceofsummer:
Duplicate of #993

PR #1128, opened 2018-03-03
Add "Remove" method to RecurringJobManager

ColinM9991:
Extends the `IRecurringJobManager` interface to provide a `Remove` method which will throw an exception of type `JobNotFoundException` if the job is not found.

The reason I have made this change is because the system will not fail fast or report any errors if you attempt to delete a job which doesn't exist which could be due to an invalid ID being stored in the database which should be mapped to a job.


ColinM9991:
Abandoning this because the pull request has been polluted by dozens of other commits.

Is there any reason for this pollution?

odinserj:
Oh, it's my bad experience in command-line client, sorry for this 😢 Regarding the original commit, could you describe more about your case, what type of recurring jobs do you have, why do you want to get an exception, etc. Basically, any interface changes are breaking changes, and aren't allowed in 1.X.

ColinM9991:
I can recreate if allowed at some point without it being a breaking change.

So what I have is an ASP.NET Core application hosted via Topshelf which uses Hangfire to execute jobs. Recently I done some work which corrupted the job ID's as they're being returned from the API to the client to link the scheduled job with the owning client.

The issue here is that because the `RemoveIfExists` method won't report if the job you've requested to remove doesn't exist, whereas I think it would be good functionality to have for logging purposes so that if a client on the front end requests to delete a scheduled recurring job, with a job ID from the front end database that doesn't exist in Hangfire, then I would be able to log this error and know quicker if there is a bug or issue with the software.

For the meantime I have implemented this as an extension method.

PR #1144, opened 2018-03-19
Stop using app locks in transaction to prevent deadlocks

odinserj:
After restructuring clustered indexes (see PR #898) we can avoid deadlocks by simply sorting our commands by their key values, leaving lock acquisitions to be always sorted. This results in improved concurrency, when disjoint set of keys is updated, since no global locks are used anymore.

Application locks are still used to serialize access to jobs, recurring jobs and batches.

### DIFF
--- a/src/Hangfire.SqlServer/SqlServerWriteOnlyTransaction.cs
+++ b/src/Hangfire.SqlServer/SqlServerWriteOnlyTransaction.cs
@@ -34,13 +34,16 @@ namespace Hangfire.SqlServer
     {
         private readonly Queue<Action<DbConnection, DbTransaction>> _queueCommandQueue
             = new Queue<Action<DbConnection, DbTransaction>>();
-        private readonly Queue<Tuple<string, SqlParameter[]>> _commandQueue
-            = new Queue<Tuple<string, SqlParameter[]>>();
         private readonly Queue<Action> _afterCommitCommandQueue = new Queue<Action>();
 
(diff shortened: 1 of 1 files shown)


AdemCatamak:
Hi,

I would like to ask a question. My app uses Hangfire with MsSql and in one day, 6 million jobs are triggered with it. Some times jobs enqueue in short amount of time and this situation leads to deadlock in database. Does this pr mittigate or solve my problem? If it does, when do you plan to publish Hangfire 1.7.0 version.

Have a good day.

odinserj:
Hi @AdemCatamak. Would be happy to know the queries that lead to deadlock. Current version shouldn't cause them, as coarse-grained locking is used to eliminate them. If there are deadlocks in Hangfire itself, it is a bug.

AdemCatamak:
Hi @odinserj, 

When I get Db reports, I see queries which are below in cause of deadlock. By the way, I use Hangfire 1.6.13 version and application runs at two machines.

You say deadlock should not be occered current version. I wonder, was this issue solved later than 1.6.13?

`(@count int,@now datetime)set transaction isolation level read committed; delete top (@count) from [HangFire].[Job] with (readpast) where ExpireAt < @now option (loop join, optimize for (@count = 20000)); at unknown line 1 (@count int,@now datetime)set transaction isolation level read committed; delete top (@count) from [HangFire].[Job] with (readpast) where ExpireAt < @now option (loop join, optimize for (@count = 20000));`

PR #1150, opened 2018-03-27
Let's try Pull Request #755 again...

eprod-edblakes:
The following is from pull request #755. The changes made are from eprod-rhansen. Mr. Hansen is a consultant that made the majority of the commits on behalf of my employer. I'm picking up this work because we would like to upgrade to the latest Hangfire but really need the upgrades made by Mr. Hansen.

From original pull request:

In working with this product in a distributed architecture, we have noticed our logs becomming inundated with job load exceptions from servers that are trying to schedule jobs when they do not have a reference to the assembly that contains the type that contains the job's logic.

This series of changes aims to isolate job scheduling such that servers only attempt to schedule or enqueue jobs that are assigned to a queue that the server is responsible for. Additionally, schedulers will now lock at the queue level which allows mutliple schedulers to run simultaneously so long as they are working from different queues.

None of these changes are breaking and can roll into any environment that is currently using the version that this change set was branched from.


burningice2866:
To make this code fully backward compatible you need to add explicit overloads to the `BackgroundJob` class since introducing method arguments with optional attributes will cause `MethodNotFound` exceptions in existing code that was using that method, without the newly added optional argument.

This is due to the nature of optional arguments being a compiler-trick and not something that is known by the runtime.

Frustrating, yes.

pieceofsummer:
This is a breaking change. You add `AddToSetQueue` methods to `IWriteOnlyTransaction`, `GetAllItemsFromSetQueue` to `JobStorageConnection` etc.

burningice2866:
In regards to the interfaces mentioned by @pieceofsummer will you need to add those new methods to a new explicit interface, ie. `IQueueWriteOnlyTransaction` and `IQueueAwareStorageConnection` and implement them - that way you make it optional for 3rd party implementers to use those interfaces.

Interfaces are a pain in that regards.

`JobStorageConnection` and `JobStorageTransaction ` would also need to have a default implementation of those newly added methods, not just marking them as abstract, to not break existing 3rd party code that inherit from those classes.

You would prob. also need to make a few `as` conversion checks to only call the new Queue-aware methods if the underlying connection or transaction is actually implementing `IQueueWriteOnlyTransaction` and `IQueueAwareStorageConnection`

burningice2866:
Version 1.7.0 has breaking changes though to ie. ´ JobStorageConnection´ so i would aim at rebasing your changeset to [1.7.0-beta1](https://github.com/HangfireIO/Hangfire/commit/5ab7e98221b4d31480dd6c105d86635694f41781)

src/Hangfire.Core/Storage/JobStorageTransaction.cs
```diff
@@ -88,5 +86,15 @@ public virtual void RemoveSet([NotNull] string key)
         {
             throw new NotSupportedException();
         }
+
+        public virtual void AddToSetQueue(string key, string value, string queueName)
+        {
+            throw new NotSupportedException();
```
pieceofsummer: Any `JobStorageTransaction` subclass is technically an `IQueueWriteOnlyTransaction`, so the code will always call `AddToSetQueue`. This will break all 3rd-party storage implementations that inherit their transaction class from `JobStorageTransaction`.

src/Hangfire.Core/Storage/JobStorageTransaction.cs
```diff
@@ -88,5 +86,15 @@ public virtual void RemoveSet([NotNull] string key)
...
+        }
+
+        public virtual void AddToSetQueue(string key, string value, string queueName, double score)
+        {
+            throw new NotSupportedException();
```
pieceofsummer: Same here.

src/Hangfire.Core/Storage/JobStorageConnection.cs
```diff
@@ -23,7 +23,7 @@
 
 namespace Hangfire.Storage
 {
-    public abstract class JobStorageConnection : IStorageConnection
+    public abstract class JobStorageConnection : IQueueStorageConnection
```
pieceofsummer: There're still new abstract methods in `JobStorageConnection` which will break all existing implementations.

burningice2866:
Breaking changes are hard :/ 

@pieceofsummer @odinserj is there any plans of making a 2.0 branch in the near future were improvements like this one, which causes breaking changes, can be merged unto?

src/Hangfire.Core/Storage/JobStorageTransaction.cs
burningice2866: You're right and IMO that's a reason why these JobStorageTransaction and JobStorageConnection should have been marked as internal, since 3rd party implementors should implement the interface and not these abstract classes.

The same could be said about a commit like this though? https://github.com/eprod-edblakes/Hangfire/commit/9b3c768dbec4afdd48a488ae0e6f173c9579258a#diff-2310048f57f3c445669c492df0f5e33b

pieceofsummer:
@eprod-edblakes thanks for your effort. Unfortunately, this is unlikely to be merged, sorry.

We've discussed the problem with @odinserj yesterday, and we have another approach to microservice architecture in mind, hopefully without breaking all the existing storages. This won't happen very soon (at least, not in 1.7.0), as there's a lot of things to consider here, but it is definitely in the backlog.

PR #1151, opened 2018-03-27
Fix compatibility with Newtonsoft.Json 11 / netstandard2.0

### LINKED ISSUE #1135: Incompatibility with Newtonsoft.Json 11.0.1 / netstandard2.0
Newtonsoft.Json 11.0.1 adds a separate target for netstandard2.0, so .NET Core 2.0 projects will use it. 

But Hangfire is built against a netstandard1.3 version. There's no built-in `FormatterAssemblyStyle` type in netstandard1.3, so Newtonsoft.Json defines one. If using Hangfire in a netstandard2.0 project, the following setter won't be resolved since there's no `FormatterAssemblyStyle` type in Newtonsoft.Json 11.0.1 anymore.

https://github.com/HangfireIO/Hangfire/blob/662e5b583a5bf34c323603ccb4258146b85f10a4/src/Hangfire.Core/States/AwaitingState.cs#L211

To fix this, Hangfire also needs a new target for netstandard2.0. Using a newer `FormatterAssemblyStyleHandling` property is not an option, since it would require updating Newtonsoft.Json dependency (it was first introduced in 10.0.1).

pieceofsummer:
Fixes #1135


odinserj:
Yay, finally merged this PR. Thanks @pieceofsummer!

PR #1152, opened 2018-03-27
Provide IntValue for all metrics

### LINKED ISSUE #1145: Missing intValue-values in /stats-endpoint
The only metric that does seem to get its `intValue` set is the `succeeded`-property.

Results from endpoint

Version used in screenshot is `1.6.17`

pieceofsummer:
Fixes #1145


odinserj:
Looks good, merging!

Issue #1071, opened 2017-12-08
Jobs retry even when automatic retry is disabled if thread aborts due to error


mikeholczer:
We've had some cases where code in one of our jobs is causing the job's thread to abort either due to a OutOfMemoryException, StackOverflowException or in trying to debug this issue by explictity calling CurrentThread.Abort(). When this happens, even though we have `GlobalJobFilters.Filters.Add(new AutomaticRetryAttribute { Attempts = 0 });` the jobs logs the following and repeatable attempts to retry the job until all threads have been killed.

`Error occurred during execution of 'Worker #8c566fd6' process. Execution will be retried (attempt 1 of 2147483647) in 00:00:01 seconds.`

parekhkb:
I've seen this behavior as well.

PR #1155, opened 2018-03-30
Adding ability for a job to throw a Non-Retryable exception type

parekhkb:
- Added a NonRetryableException type
- Added support in AutomaticRetryAttribute.cs to prevent retries if a failed job has a exception of type NonRetryableException


odinserj:
Could you add some details regarding the problem you are facing with and why it can't be solved using the existing functionality?

src/Hangfire.Core/Hangfire.Core.csproj
```diff
@@ -93,6 +93,7 @@
     <Compile Include="Dashboard\Owin\OwinDashboardContextExtensions.cs" />
     <Compile Include="Dashboard\Owin\OwinDashboardRequest.cs" />
     <Compile Include="Dashboard\Owin\OwinDashboardResponse.cs" />
+    <Compile Include="NonRetryableException.cs" />
```
pieceofsummer: Also needs to be included in Hangfire.Core.NetStandard.csproj

pieceofsummer:
@odinserj this actually makes sense, since some exceptions shouldn't be retried. For example, the job argument was invalid, or there's some configuration error. Further retries won't make it valid, so there's no reason to keep the job failing again and again.

odinserj:
I'm just willing to add some more details to the PR 😊 Also, there's another implementation of this feature without introducing the new classes with the following API (don't consider property name too seriously):

```csharp
[AutomaticRetry(BlackList = new [] { typeof(InvalidOperationException) })]
```

parekhkb:
We have some cases where some exceptions should not be retried. An example would be that a duplicate primary key exception is encountered when trying to commit to a DB. Such an exception will never succeed no matter how many times you retry. 

I will explore adding a black-list parameter to the constructor.

odinserj:
Thanks for giving the details. I've just proposed the changes, and you will be the first user of this feature, so if you have some argument against it – just let us know!

parekhkb:
@odinserj Thanks! Do you need an Issue to track this? Or do you plan to take in this PR?

src/Hangfire.Core/AutomaticRetryAttribute.cs
```diff
@@ -101,6 +108,26 @@ public AutomaticRetryAttribute()
             Order = 20;
         }
 
+        /// <summary>
+        /// Gets or sets a list of Non-retryable exception types.
+        /// </summary>
+        /// <value>Any non-negative number.</value>
+        /// <exception cref="ArgumentOutOfRangeException">The value in a set operation is less than zero.</exception>
+
+        public Type[] NonRetryable
```
odinserj: Looks like there's another word for this – https://en.wiktionary.org/wiki/retriable 😊

parekhkb:
@odinserj, Your earlier comments about proposing the changes. Do you track those somewhere privately? I'm just looking for a ballpark of when to expect the changes. 

Also, I've hit a snag on this PR. the context is never a FailedState, even if I expliclty throw an exception in my job. This is the issue and it might be blocking these changes: https://github.com/HangfireIO/Hangfire/issues/1071

PR #1158, opened 2018-04-02
enable fast scheduler wakeup

pm100:
The polling delay on the scheduler is done in a different way from the queue poll delay. This means that the scheduler wait cannot be preempted by an event (the queue wait is in the storage plugin so the implementer can wake up anyway they like)

The change allows the hoster to provide an event that the scheduler should wait on

### DIFF
--- a/src/Hangfire.Core/BackgroundJobServer.cs
+++ b/src/Hangfire.Core/BackgroundJobServer.cs
@@ -162,7 +162,7 @@ namespace Hangfire
                 processes.Add(new Worker(_options.Queues, performer, stateChanger));
             }
             
-            processes.Add(new DelayedJobScheduler(_options.SchedulePollingInterval, stateChanger));
+            processes.Add(new DelayedJobScheduler(_options.SchedulePollingInterval,_options.SchedulerNotificationEvent, stateChanger));
             processes.Add(new RecurringJobScheduler(factory));
 
             return processes;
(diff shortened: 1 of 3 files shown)


pieceofsummer:
1. This PR changes signature of a public constructor, hence is a breaking change.

2. It won’t work in distributed scenarios.

3. I don’t see how exactly this may be used, since delayed jobs aren’t triggered by some external event but rather by the time. Can you elaborate on the specific use case?

pm100:
1.  Will fix.
  2.  Works in distributed environment – that’s how I am using it. What you don’t see is how the notifications get from machine to machine. I do that using my own plumbing and then set the event
  3.  A job gets inserted into the job table by a client. This is used to trigger this event, the code that does that is in my plugin. Just the same way you do the queue wakeup in the sqlserver plugin. Works almost exactly the same.

The thing you don’t see is that in the plugin that stores a row in the job table for the client I then broadcast to other machines.
I already have that working on the queue wakeup in the plugin, but the way the scheduler polls work means that the equivalent wakeup cannot be done

Eg

In my AddToSet

                    if (key == "schedule")
                    {
                        if (score == 0.0
                        || score < JobHelper.ToTimestamp(DateTime.UtcNow + TimeSpan.FromSeconds(10))) // XXX need real scheduler timer
                        {
                            // we just scheduled a job that starts in < the scehduler watchdog timer
                            // poke scheduler
                            CloudFireRpc.NotifyScheduledJob();
                        }
                    }

That NotifyScheuledJob uses my machine to machine broadcast RPC. The receiving code sets the event that the scheduler is waiting on

Use case is that I submit a job to execute right now, it gets run immediately. Without having ultra aggressive timers

Sent from Mail<https://go.microsoft.com/fwlink/?LinkId=550986> for Windows 10

________________________________

pm100:
Ok – I did it a different way. Ty for the pointer

Sent from Mail<https://go.microsoft.com/fwlink/?LinkId=550986> for Windows 10

________________________________

PR #1177, opened 2018-05-04
Use strong name

quinvit:
Applying strong name for Hangfire.


odinserj:
Hi @quinvit. Have you tried to use the [`strongnamer`](https://github.com/dsplaisted/strongnamer) project that allows you to transparently handle non-signed projects?

caioproiete:
@odinserj [StrongNamer](https://github.com/dsplaisted/strongnamer) only works for top-level apps that need to be strong-name signed and depend on libraries that are not yet signed.

For libraries that can be added as dependencies of other libraries (such as `Hangfire.Core` for example) it doesn't work, because the key used to sign matters in the dependency chain.

e.g.:
* Paul creates `LibraryA` that depends on `Hangfire.Core` v1.7.6 and use [StrongNamer](https://github.com/dsplaisted/strongnamer) with key `X`
* Mary creates `LibraryB` that depends on `Hangfire.Core` v1.7.6 and use [StrongNamer](https://github.com/dsplaisted/strongnamer) with key `Y`

Now we are in a bad place... Both `LibraryA` and `LibraryB` seem to depend on the exact same assembly, yet for the .NET Framework they are completely different assemblies because the key used to sign them is part of the identity...

PR #1190, opened 2018-05-25
Added Queue column to RecurringJobsPage

### LINKED ISSUE #1189: Recurring jobs on the dashboard don't show which queue it is using
The "Recurring Jobs" page of the Hangfire Dashboard does not show which queue it is creating the jobs on.

2018-05-24 14_27_44-recurring jobs - hangfire

Deantwo:
This is untested code, but should work in theory.

Fixes #1189

### DIFF
--- a/src/Hangfire.Core/Dashboard/Content/resx/Strings.resx
+++ b/src/Hangfire.Core/Dashboard/Content/resx/Strings.resx
@@ -343,6 +343,9 @@
   <data name="RecurringJobsPage_Table_Cron" xml:space="preserve">
     <value>Cron</value>
   </data>
+  <data name="RecurringJobsPage_Table_Queue" xml:space="preserve">
+    <value>Queue</value>
+  </data>
   <data name="RecurringJobsPage_Table_LastExecution" xml:space="preserve">
     <value>Last execution</value>
   </data>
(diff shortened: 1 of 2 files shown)


pieceofsummer:
Thanks for your PR! Would you mind to commit updated `Strings.Designer.cs` and also provide localized strings for other languages (you can simply copy them from corresponding QueuesPage_Table_Queue)?

Deantwo:
I think that should do it.

odinserj:
@pieceofsummer, merging? For 1.6.20 (`master` branch) we should ensure that the dashboard page looks fine, for 1.7.0 (`dev` branch) we can postpone this task to test it later.

pieceofsummer:
@odinserj this shouldn't break anything, but it is not an urgent change.

We might first merge it into `dev` (and release the second beta), then merge into `master` if there are no layout issues because of increased content width.

odinserj:
@pieceofsummer, sounds good!

pieceofsummer:
@Deantwo unfortunately, it won’t be as easy. There’s no Queue field associated with the Processing state, so it would require either adding it, or extracting from the previous Enqueued state.

Deantwo:
@pieceofsummer nevermind then. Issue just seemed very similar.

Deantwo:
So, is this being merged?

Deantwo:
What changed to make those tests fail?
This is a very simple little change, so getting it merged would be nice.

I was even missing this feature yesterday.

Deantwo:
This isn't getting merged because it would make it more obvious there is a big issue with recurring jobs and queues?
Since I just ran head first into the #595 issue myself, and this minor change would make that a lot more obvious.

PR #1193, opened 2018-05-30
Move state change retries to the Worker class

odinserj:
There was a change 8570ef77cc349c30ea7d105197aeb0f5d406d7c6 in Hangfire 1.6.12 that added retries for a state changing process in case of exceptions. This was a good fix for handling errors in buggy state filters, but there are two major problems with the implementation:

1. When all retry attempts exhausted, we were applying the `Failed` state by using a "lightweight" state machine that isn't invoking any job filters, resulting in a rare, but possible non-consistent behavior, especially when there's a storage outage (that's recovered just after the last retry attempt), for all the filters that are based on the "previous" state.
2. An exception may be caused by a broken connection. In this case, distributed lock acquired in the `ChangeState` method may be released, if lock implementation is based on a connection. And all the retries will be invoked without any attempt to restore the lock. Luckily, all the further attempts will cause an exception in Hangfire.SqlServer 1.6.17, but for other storages there may be a problem (big problem).

With the proposed change, the retry logic is moved to the `Worker` class instead and changes `BackgroundJobStateChanger` to immediately throw an exception in case of an error. Worker is catching these exceptions and retries the state changing process from scratch with proper acquisition of all the related locks.

When all the retry attempts are exhausted, worker will attempt to move the background job to the failed state, using the full-featured state machine with all the required filters. When the latter step gives us an exception, our worker will report it as an error and place the identifier back to its queue, allowing other workers to repeat this long process (since there are delays during the retries) full of warning messages.

### DIFF
--- a/src/Hangfire.Core/Server/Worker.cs
+++ b/src/Hangfire.Core/Server/Worker.cs
@@ -41,6 +41,8 @@ namespace Hangfire.Server
     public class Worker : IBackgroundProcess
     {
         private static readonly TimeSpan JobInitializationWaitTimeout = TimeSpan.FromMinutes(1);
+        private static readonly int MaxStateChangeAttempts = 10;
+
         private static readonly ILog Logger = LogProvider.For<Worker>();
 
         private readonly string _workerId;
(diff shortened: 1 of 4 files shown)


pieceofsummer:
Seems legit to me

PR #1207, opened 2018-06-22
changed Less details to Fewer details

gareth-evans:


odinserj:
Thanks @gareth-evans!

PR #1222, opened 2018-07-12
Update SqlServerObjectsInstaller.cs

TimSQL:
Support Azure Sql Managed instance


odinserj:
@TimSQL, may be we should remove that method completely?

TimSQL:
Ok, do i make another commit for that?

odinserj:
@TimSQL I've accepted this PR, because it solves the issue anyway. If you wish to remove that method completely, I will be happy to accept the new PR as well. Thanks!

PR #1230, opened 2018-07-17
Update SqlServerObjectsInstaller.cs

TimSQL:
In addition #1222: Removed the function IsSqlEditionSupported to support future versions of SQL Server.


odinserj:
Thanks @TimSQL! Deleting lines is the most awesome stuff in programming 😃

PR #1242, opened 2018-08-06
Fix for #861 Dashboard background polling does not handle 401 responses

### LINKED ISSUE #861: Dashboard background polling does not handle 401 responses
We have implemented our own authentication using OpenID connect. Everything works fine except:
When authentication session is lost (could not refresh auth token because e.g. was revoked) the dashboard does not correctly handle "HTTP 401" responses when polling .. it simply keeps polling forever.

prochnowc:
- Fixes #861 
- Implemented error popup when polling fails.
- Stop polling when request failed
- Use setTimeout() instead of setInterval() when polling to not outperform pending requests


prochnowc:
@odinserj what do you think?

prochnowc:
@odinserj Ping

odinserj:
Thanks for your work @prochnowc and sorry for waiting for so long. I've modified the behavior a bit by showing an alert box instead of a modal window, and also added automatic page refresh attempt that triggers after a minute.

PR #1246, opened 2018-08-07
Add Result property to StateData, fixes #1245

### LINKED ISSUE #1245: Add a Result property to Job or JobData class
I'm using this code to check whether a job has completed successfully or not.
```
using (var conn = JobStorage.Current.GetConnection()) {
  var data = conn.GetJobData(Id);
  if (data?.State == SucceededState.StateName) {
    //Here, I need to get the result
  }
}
```
When a job is successfully completed, I need to get its result but there's no Result property in the JobData or Job classes. What I'm doing, instead, is the following, which I think is not efficient.
```
var succeededJob = JobStorage.Current.GetMonitoringApi().SucceededJobs(0, int.MaxValue).FirstOrDefault(j => j.Key == Id);
var jobResult = succeededJob.Value.Result;
```

How can I efficiently get a job result by its job id?

BrightSoul:


pieceofsummer:
Result is only valid for Succeeded state, and is irrelevant for all other states. So it doesn’t make a lot of sense to add it as a property here.

Issue #1159, opened 2018-04-04
BackgroundJob.Schedule - queue name?


drauch:
Hi!

Normally we schedule our jobs using `BackgroundJob.Enqueue(methodCall)` to a specific queue. One of those jobs should have a delay before being executed - so we used `BackgroundJob.Schedule(methodCall, delay)`. However, it looks like such jobs are not scheduled for a specific queue, it looks like any worker process will happily try to execute the job as soon as the delay has passed. That's a problem for us, as not any worker has loaded the necessary DLLs to execute the job, therefore, the job often fails! Is there a way to make sure that the job is only given to specific types of workers - i.e., the workers attached to the queue we scheduled the job on?

My guess is, that `BackgroundJob.Schedule` doesn't schedule a job for a specific queue by default. How to configure it so that the job is scheduled for the given queue?

Best regards,
D.R.

drauch:
OK, it looks like the other workers are not trying to execute the job, they are only trying to move the job from scheduled to enqueued - however, they still need all the DLLs required for that job for doing this...

...is there a way to stop workers from trying to move scheduled jobs to the enqueued state? So that only our "server" is trying to do that? Any other workaround?

drauch:
We looked a bit into the code, and it looks like BackgroundJobServer "hardcoded" starts a DelayedJobScheduler and we cannot disable it in any way. Would you accept a PR that makes `BackgroundJobServer.GetRequiredServices()` virtual and would allow us to disable the `DelayedJobScheduler`?

Then we could use a derived BackgroundJobServer in all workers which are not having references to all the other worker assemblies and run only a specific worker with a DelayedJobScheduler.

virtualcca:
i also have same problem. i find DelayedJobScheduler enqueue is not special queueName so it also is "Default"

drauch:
Is this issue tracker alive?

christogav:
Need this! We've just hit this issue... how far away is a fix?

drauch:
@christogav : Unfortunately, it looks like the project is not actively maintained. I've never heard back from them, most issues just sit there and rot :-(

christogav:
Is #1247 the fix for this? Appears related.

Frank591:
Our team also need this. Pls fix

zhuweid:
same here

odinserj:
[...]

PR #1247, opened 2018-08-13
Scheduling to specific queue

cdschneider:
Although it looks like it's been partially attempted in the past, solving #748 (providing support for scheduling jobs to a non-default queue).

This change includes:

* An added `CandidateQueue` property to `ScheduledState` (that also gets serialized/persisted w `Job`s `StateData`
* A new element in the `ScheduledState`s details section that includes a link to the actual queue's HF details page
* Optional `queue` parameter to `BackgroundJobClientExtensions` on all `Schedule` methods

Here is an example of the scheduling with this feature:

```
BackgroundJob.Schedule(
    () => Console.WriteLine("Hello, world!"),
    TimeSpan.FromHours(1), 
    "custom_queue");
```

HF job details when in "Scheduled" state looks like the following:

image


Nexsus94:
Any updates on this?

pieceofsummer:
@cdschneider thanks for your work.

This would require a careful review before merging. But adding arguments with default value is absolutely a no go, as it breaks binary compatibility. Would you please change them into the corresponding methods' overloads instead?

drauch:
Is there any work going on regarding this pull request? It would probably fix our issue #1159, which is why we're interested in it. Ping @cdschneider

cdschneider:
@drauch sorry for delay on this. @pieceofsummer I've refreshed my branch from `Hangfire:dev` and updated per your comments to remove default-parameters and made the appropriate `Schedule` method overloads in the `BackgroundJobClientExtensions` class. 

Please let me know what you think and if you'd like to see more updates prior to merging.

weyert:
How can we progress this PR?

willemduncan:
Would like to see this feature as well. Don't want to hard code queue names in a function attribute, the chosen queue will depend on runtime variables which are only know during enqueue.

willemduncan:
We run a filter that writes the queue name as job parameter, and then uses that to overwrite whatever hangfire decides should be the queue later. It's globally applied at app startup:

```cs
GlobalJobFilters.Filters.Add(new PreserveOriginalQueueFilter());
```

And looks like this:

```cs
/// <summary>
/// Job filter that preserves the job queue when changing state to 'Enqueued'.
/// </summary>
public class PreserveOriginalQueueFilter : IApplyStateFilter
{
    private const string ParameterName = "OriginalQueue";

    {
        // Activating only when enqueueing a background job.
        if (context.NewState is EnqueuedState enqueuedState)
        {
            // Checking if an original queue is already set.
            var originalQueue = SerializationHelper.Deserialize<string>(context.Connection.GetJobParameter(context.BackgroundJob.Id, ParameterName));
            if (originalQueue != null)
            {
                // Override any other queue value that is currently set.
                enqueuedState.Queue = originalQueue;
            }
            else
            {
                // Queueing for the first time, we should set the original queue.
                context.Connection.SetJobParameter(
                    context.BackgroundJob.Id,
                    ParameterName,
                    SerializationHelper.Serialize(enqueuedState.Queue)
                );
            }
        }
    }

    {
    }
}
```

frankyjquintero:
@cdschneider  @odinserj What do we need to add this functionality, its use simplify dependencies in cleaner architecture

odinserj:
Thanks for your patience. Upcoming 1.8.0 release will bring first-class support for custom default queues for background jobs so almost nothing is required:

```csharp
BackgroundJob.Schedule("critical", () => Console.WriteLine("Hi!"), TimeSpan.FromHours(1));
```

Queue is persisted with background job information itself so everything is preserved on automatic retries and other state changes. This feature is already available in the latest betas and will be released approximately next week.

frankyjquintero:
Excellent news, I am looking forward to this new version. Thank you

fbonjour81:
> Thanks for your patience. Upcoming 1.8.0 release will bring first-class support for custom default queues for background jobs so almost nothing is required:
> 
> ```cs
> BackgroundJob.Schedule("critical", () => Console.WriteLine("Hi!"), TimeSpan.FromHours(1));
> ```
> 
> Queue is persisted with background job information itself so everything is preserved on automatic retries and other state changes. This feature is already available in the latest betas and will be released approximately next week.

Great news @odinserj!!! Could we have some kind of conflict using QueueAttribute and PreserveOriginalQueueAttribute with this new version? The idea is remove them in the future

Issue #928, opened 2017-07-06
Hangfire 1.7.0 roadmap


odinserj:
### Project

* [x] Migrate to the new csproj project format, remove compatibility dependency (#982, PR #1309).
* [ ] ~~Move Hangfire.AspNet package to the main repository and document it.~~ Later.
* [x] Add `netstandard2.0` and `net46` targets to all the projects.
* [x] Internalize Cronos dependency on `netstandard` targets using ILRepack.
* [ ] ~~Add SourceLink support to improve debugging experience.~~ Later.

### Background processing

* [x] Add support for `Task`-based async background processes, opens road towards async storage (Issue #150).
* [x] Use circuit breaker pattern to reduce the number of log messages in case of a failure.
* [x] Background processing server should restart itself with a new id after server timeout to avoid resurrections (Issue #796).
* [x] Add ability to stop workers from picking new jobs without cancelling processing of current ones (Issues #225, #402).
* [x] Re-consider retries: add ability to use custom delays.

### Storage

* [x] Decrease the size of serialized payloads and remove version information (PR #891).
* [x] Convert identity types to `bigint` in Hangfire.SqlServer to remove upper limit for job count (PR #898, Issue #749).
* [x] Re-consider indexes in `SqlServerStorage`: remove unnecessary, create neccessary ones (PR #898, Issues #658, #844).
* [ ] ~~Add built-in index maintenance for database objects in Hangfire.SqlServer.~~ This should be documented instead.
* [x] Make transactionless fetch introduced in Hangfire.SqlServer 1.6.14 the default one.
* [x] 1.6.X workers should be able to handle 1.7.0 schema and vice versa.
* [x] Make Hangfire.SqlServer 1.7.0 migration optional, because it may take some time.
* [x] Document gradual migration to 1.7.0 step-by-step.
* [x] Add the `GetByLowestScoreFromSet(..., int count)` overload (#38).
* [x] Implement a long-polling based on short-lived application locks.

### Recurring jobs

* [x] Support for complex Cron expressions, like `L`, `W`, `#` characters (PR #853, Issue #494).
* [x] Solve problems with Daylight saving time transitions (PR #853, Issue #215).
* [x] Handle different time zone names between Windows and Linux.

### Misc

* [x] Add DisplayNameAttribute-like feature when running in .NET Core (Issues #635, #718).
* [x] ~~Change TransactionTimeout default value.~~
* [ ] ~~Change default value for polling interval in dashboard.~~
[...]

Issue #38, opened 2014-03-05
Improve scheduled jobs performance


odinserj:
Scheduled jobs are fetched one by one now. Despite of Redis implementation executes much better than SQL Server one, this task can be performed better.
To do this, schedule poller should take top N jobs that need to be enqueued. To prevent unnecessary locks in a multi-server environment, we could shuffle this list. So each server will take random job to enqueue.

odinserj:
Scheduled jobs implementation is the bottleneck of delayed and recurring jobs. There's no simple way to run our `DelayedJobScheduler` concurrently without introducing new storage abstractions (and it already have a lot of them). The solution proposed in the first comment by shuffling the list will not work great, because there may be contention anyway, and more importantly is a very hard operation to implement in a storage itself.

Instead, a new method `GetByLowestScoreFromSet(..., int count)` will be added in 1.7.0 for the connection class that allows us to take multiple scheduled job identifiers at once, and we could batch multiple state transitions together. This will reduce the number of round-trips required. 

After performing this step, it will be possible to implement new states for delayed jobs to not to break anything, but this will be another story.

PR #1251, opened 2018-08-22
Adding GetFirstByLowestScoreFromSet overload

cdschneider:
First step in improving performance of `DelayedJobScheduler` background process, as mentioned in #38. This is also listed in the [Hangfire 1.7.0 roadmap](https://github.com/HangfireIO/Hangfire/issues/928).

The next step following this feature is to implement batch state transitions for the batch of retrieved jobs to be scheduled.


odinserj:
Thanks @cdschneider for implementing this and writing tests! I've removed breaking changes as proposed by @burningice2866 (nice catch!) and did some cosmetic changes. Now waiting on CI results and merging it as soon as they are ready.

I've also did some tests using this new method and rewritten `DelayedJobScheduler`, and seen up to 70% better throughput when using delayed jobs using SQL Server.

PR #1274, opened 2018-09-27
Allow `Type` to be passed down.

BardezAnAvatar:
The functionality is already present, but only present for Genetics.


BardezAnAvatar:
Due to sloppy nature of tacking on the `Type`, I am closing this PR and replacing it with a different approach of additional extension methods

PR #1278, opened 2018-10-01
SQL Storage: Dequeue a job based on queue priority

jerms0518:
This pull request adds the ability to define a queue priority based on the array index of the queue in the `BackgroundJobServerOptions` object.

For example, a `BackgroundJobServer` set up with this queue definition will process items from the "default" queue first and then the "critical" queue if there are no jobs remaining in "default":

```C#
BackgroundJobServerOptions options = new BackgroundJobServerOptions();
options.Queues = new[] { "default", "critical" };
```

To set the "critical" queue as top priority for a `BackgroundJobServer`, this configuration can be used:

```C#
BackgroundJobServerOptions options = new BackgroundJobServerOptions();
options.Queues = new[] { "critical", "default" };
```

The current functionality dequeues items from queues based on queue name instead of order in the array, so this change adds flexibility to process queues in different orders on different servers.

### DIFF
--- a/src/Hangfire.SqlServer/SqlServerJobQueue.cs
+++ b/src/Hangfire.SqlServer/SqlServerJobQueue.cs
@@ -131,11 +131,28 @@ where Queue in @queues and
             FetchedJob fetchedJob = null;
             DbTransaction transaction = null;
 
+            string queueNames = "";
+            for (int i = 0; i < queues.Length; i++)
+            {
+                var queueName = queues[i];
+
+                if (i > 0)
+                {
+                    queueNames += ", ";
+                }
+
+                queueNames += $"('{queueName}', {i + 1})";
+            }
+
             string fetchJobSqlTemplate =
-                $@"delete top (1) JQ
+$@"delete from [{_storage.SchemaName}].JobQueue
 output DELETED.Id, DELETED.JobId, DELETED.Queue
+where Id in (
+select top (1) Id
 from [{_storage.SchemaName}].JobQueue JQ with (readpast, updlock, rowlock, forceseek)
-where Queue in @queues and (FetchedAt is null or FetchedAt < DATEADD(second, @timeout, GETUTCDATE()))";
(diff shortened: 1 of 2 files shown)


src/Hangfire.SqlServer/SqlServerJobQueue.cs
```diff
@@ -131,11 +131,51 @@ private SqlServerTransactionJob DequeueUsingTransaction(string[] queues, Cancell
...
+  begin
+    select * from @output
+    break
+  end
+
+  if (@index > {queues.Length})
+  begin
+    select * from @output
+  end
+end";
```
pieceofsummer: Could be just this:
```c#
                string fetchJobSqlTemplate = $@"
with JQ as (
    select top (1) JQ.*
    from [{_storage.SchemaName}].JobQueue JQ with(readpast, updlock, rowlock, forceseek)
        inner join (values {string.Join(", ", queues.Select((q, i) => $"('{q}', {i})"))}) Q(Queue, Priority)
            on Q.Queue = JQ.Queue
    where (FetchedAt is null or FetchedAt < DATEADD(second, @timeout, GETUTCDATE()))
    order by Q.Priority
)
delete from JQ
output DELETED.Id, DELETED.JobId, DELETED.Queue";
```

More importantly, you're introducing a classical SQL injection vulnerability here. Queue names may be coming from untrusted sources (e.g. configuration file), and should be properly escaped before including them into the query.

P.S. To be fair, the original queries are already vulnerable via `_storage.SchemaName`, but it's not a good enough reason to screw things up even more :)

jerms0518: Good point about the SQL Injection vulnerability. I unfortunately don't have access to the code I was working on when I made these changes, but I remember running into issues when using a simpler join with an `order by` like in your suggestion because the order causes all rows to be locked when used with the specified SQL hints (`updlock`, `rowlock`, etc.). This was hindering the parallel processing because jobs needed to complete before the rows would be released for the next server to process. Have you seen similar problems?

Issue #1295, opened 2018-11-08
Value defining max length for parameters should be configurable by the client


emanuelmarques:
Greetings,

With the last update, we stoped being able to see the parameters of most of our jobs. We checked the line bellow and the max length is hardcoded. This value should be configurable by whoever uses hangfire.

Thank you

https://github.com/HangfireIO/Hangfire/blob/053afff4b7fcfa2be060760b14879c5ac7a93204/src/Hangfire.Core/Dashboard/JobMethodCallRenderer.cs#L32

odinserj:
I understand that hard-coded values are bad, but could you add more details on what arguments you are using, why so large, and what number of max argument size you are planning to use?

odinserj:
@emanuelmarques, I really don't want to add yet another configuration option for this. What if I convert the `MaxArgumentToRenderSize` to a field and provide a method for changing it via reflection? In this case you will be able to change it without introducing a new option.

emanuelmarques:
@odinserj If with that, I am able to provide the max / set unlimited max argument, I'm ok with it

emanuelmarques:
Hi @odinserj 
Is there any prediction of when this is being released? or is it already released?

gmoreno90:
Hi guys,
I'm trying to use it and all the times have this error.
Cannot set initonly static field 'MaxArgumentToRenderSize' after type 'Hangfire.Dashboard.JobMethodCallRenderer' is initialized
What is the correct line where we need to change? 
I tested on Program and on startup.
Many thanks in advance
Gon

gmoreno90:
Hi @emanuelmarques this works for you?
Best regards
Gon

DotJoshJohnson:
@gmoreno90:

I just happened to be investigating the same thing today. It turns out, the ability to set readonly fields was never truly reliable and this has been explicitly disabled in .NET Core: https://github.com/dotnet/runtime/issues/11571.

I would agree with @emanuelmarques that this should be exposed as an option, so I've added my vote above.

emanuelmarques:
Hello @gmoreno90 @DotJoshJohnson 

That actually worked for me and has been working for the past almost 2 years. I may be using an older version of .net core where that ability is still possible.

However, I currently have mixed feelings about this: 
On one side, if hangfire allows big arguments to be passed to a job, it should be shown in the dashboard. 
[...]

PR #1302, opened 2018-11-14
Provide MaxArgumentToRenderSize in Dashboard options

realapak:
Fix for:
https://github.com/HangfireIO/Hangfire/issues/1295


odinserj:
Thanks for your work on this, @realapak. I've added a change to the upcoming 1.6.22 that allows to change the value without introducing a new option via reflection. Please see https://github.com/HangfireIO/Hangfire/issues/1295#issuecomment-454693246 for details.

Issue #982, opened 2017-08-30
can remove references to Microsoft .NETCore.Portable.Compatibility？


xakepbean:
use dotnet publish -r win10-x64
exec *.exe after
Failed to initialize CoreCLR, HRESULT: 0x80004005

RehanSaeed:
I'm seeing the same issue, is there a workaround?

mikeli11:
To repro:

In VS2017:
1) Create new ASP.NET application
2) In package manager console: Install-Package Hangfire
3) Add in startup - .UseHangfireServer(), .UseHangfireDashboard(), services.AddHangfire(x =>UseSqlServerStorage())
4) cd into project folder
5) in CMD: dotnet publish -c Debug -f netcoreapp2.0 -r win7-x64
6) in publish folder (\WebApplication\bin\Debug\netcoreapp2.0\win7-x64\publish) in CMD: dotnet WebApplication.dll

Error:
Failed to initialize CoreCLR, HRESULT: 0x80004005

stestet:
We are having the same problem... Is there hope?

Salazander:
We had a similar issue (Failed to initialize CoreCLR, HRESULT: 0x80004005) and fixed it by targetting runtime win-x64 instead of win10-x64. Not sure why, but it looks like the wrong version of System.Private.CoreLib.dll gets copied into the publishing folder when targetting win10-x64

graemefoster:
I'm getting this as-well when running on Cloud Foundry (dotnet publish -r ubuntu.14.04-x64)

cbmek:
It is not possible to build `win10-x64` and `win7-x64`.
Will it ever be fixed?

odinserj:
Reference to the `Microsoft.NETCore.Portable.Compatibility` package removed in the latest betas after moving the project to use new SDKs. Also, explicit `netstandard2.0` target was added to remove the reference burden so everything should be working fine now.

PR #1309, opened 2018-12-03
Migrate to the new project format

odinserj:
This PR finally migrates all the projects to the new project file format. This will help to remove the dependency on `Microsoft .NETCore.Portable.Compatibility` package that causes run-time exceptions when using published projects (#982), allow to remove all the quirks with `project.json` files and move forward with `netstandard2.0` target. The resulting project structure is much more cleaner now.

T4 text transformation works well in Visual Studio 2017 (required for `DefaultInstall.sql` script). Razor code generation also works during the build step, but only in full .NET Framework environments, as .NET Core based MSBuild utility is not enough currently. And it's great to see that Mono 5.X can compile all the new projects.

The only thing that's removed is automatic test coverage checks, but I recently heard somewhere there's a way to run coverage tests on .NET Core. Nevertheless, new tooling even without automatic coverage is much better, and I believe it's not a hard thing to be done in the near future.


gitblit:
There are several options now for .Net Core test coverage.  We've had good luck with https://github.com/tonerdo/coverlet.

PR #1319, opened 2019-01-02
Support non assembly qualified name

quinvit:
This PR resolve assembly versioning issue. Here are the steps to reproduce:

1. Register a job with any type in an assembly
2. Upgrade that assembly to new version
3. Job invocation will be result in error: cannot load assembly...

We should have an option to skip strictly on qualified name of an assembly.

### DIFF
--- a/src/Hangfire.Core/Storage/InvocationData.cs
+++ b/src/Hangfire.Core/Storage/InvocationData.cs
@@ -48,7 +48,7 @@ namespace Hangfire.Storage
         {
             get
             {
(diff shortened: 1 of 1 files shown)


pieceofsummer:
There’s already a number of changes in the `dev` branch regarding how the `InvocationData` is serialized. The upcoming version 1.7.0 already supports versionless assembly names in job definitions and also resolves system type forwarding issues between .net and .net core.

You should definitely give it a try.

Issue #1323, opened 2019-01-08
allow SQL Server to use Azure token


chinwobble:
When using Azure's managed service identity you need to keep fetching tokens that have a 1 hour expiry. 

Like this 
```
                var db = new SqlConnection(ConfigManager.ConnectionString);
                if (IsAzureDB)
                {
                    db.AccessToken = tokenProvider.GetAccessTokenAsync("https://database.windows.net/").GetAwaiter().GetResult();
                }
                return db;
```

In order to use this feature we need to extend `Hangfire.SqlServer.SqlServerStorage` to cater for this.
We can add an overload to the constructor that accepts `Func<SqlConnection>` which is called on `Hangfire.SqlServer.SqlServerStorage.CreateConnection()`

gary-walker-82:
as this is specific to Azure would it not be better suited to having a Hangfire.SqlServer.AzureSqlServerStorage class?

pieceofsummer:
This doesn't seem like a breaking change. For example, add an `Action<SqlConnection> ConnectionConfigurator` property to `SqlServerStorageOptions` and call it from `SqlServerStorage.CreateAndOpenConnection()`. We already have `ImpersonationFunc` for similar purpose in `dev` branch, so it shouldn't make too much of a difference.

@chinwobble would you like to create a pull request for this?

chinwobble:
@pieceofsummer sure I'll create a pull request for it

odinserj:
Yes, connection factory is great for this scenario, and I think we can release it within 1.6.X, since it can be critical on some environments.

PR #1325, opened 2019-01-10
Add connection factory overload for SqlServerStorage

chinwobble:
Solves #1323

### DIFF
--- a/src/Hangfire.SqlServer/SqlServerStorage.cs
+++ b/src/Hangfire.SqlServer/SqlServerStorage.cs
@@ -267,7 +267,7 @@ namespace Hangfire.SqlServer
         internal DbConnection CreateAndOpenConnection()
         {
             var connection = _existingConnection
-                ?? _connectionFactory()
+                ?? _connectionFactory?.Invoke()
                 ?? new SqlConnection(_connectionString);
 
             if (connection.State == ConnectionState.Closed)
--- a/src/Hangfire.SqlServer/SqlServerStorage.cs
+++ b/src/Hangfire.SqlServer/SqlServerStorage.cs
@@ -41,6 +41,7 @@ namespace Hangfire.SqlServer
         private readonly Func<DbConnection> _connectionFactory;
         private readonly SqlServerStorageOptions _options;
         private readonly string _connectionString;
+        private readonly string _toString;
 
         public SqlServerStorage(string nameOrConnectionString)
             : this(nameOrConnectionString, new SqlServerStorageOptions())
@@ -67,6 +68,7 @@ namespace Hangfire.SqlServer
             _connectionString = GetConnectionString(nameOrConnectionString);
             _connectionFactory = () => new SqlConnection(_connectionString);
             _options = options;
+            _toString = ToStringImpl();
 
             Initialize();
         }
@@ -93,6 +95,7 @@ namespace Hangfire.SqlServer
 
             _existingConnection = existingConnection;
             _options = options;
+            _toString = ToStringImpl();
 
             Initialize();
         }
@@ -119,6 +122,7 @@ namespace Hangfire.SqlServer
 
             _connectionFactory = connectionFactory;
             _options = options;
+            _toString = ToStringImpl();
 
             Initialize();
         }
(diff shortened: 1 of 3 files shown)


src/Hangfire.SqlServer/SqlServerStorage.cs
```diff
@@ -239,7 +266,9 @@ internal void UseTransaction(DbConnection dedicatedConnection, [InstantHandle] A
 
         internal DbConnection CreateAndOpenConnection()
         {
-            var connection = _existingConnection ?? new SqlConnection(_connectionString);
+            var connection = _existingConnection
```
odinserj: @chinwobble, thank you for the PR. I think we can simplify this method by introducing the *default connection factory* that returns a new connection based on a connection string. Otherwise we are required to add more tests to see whether `SqlServerStorage` based on a connection factory actually works.

```csharp
() => new SqlConnection(_connectionString);
```

After creating a default connection factory we can simply replace this line with the following one.

```csharp
var connection = _existingConnection ?? _connectionFactory();
```

odinserj: I think we can simplify this method by introducing the *default connection factory* that returns a new connection based on a connection string. Otherwise we are required to add more tests to see whether `SqlServerStorage` based on a connection factory actually works.

```csharp
() => new SqlConnection(_connectionString);
```

After creating a default connection factory we can simply replace this line with the following one.

```csharp
var connection = _existingConnection ?? _connectionFactory();
```

tests/Hangfire.SqlServer.Tests/SqlServerStorageFacts.cs
```diff
@@ -41,6 +42,15 @@ public void Ctor_CanCreateSqlServerStorage_WithExistingConnection()
             Assert.NotNull(storage);
         }
 
+        [Fact, CleanDatabase]
+        public void Ctor_CanCreateSqlServerStorage_WithConnectionFactory()
+        {
+            Func<DbConnection> connectionFactory = ConnectionUtils.CreateConnection;
+            var storage = new SqlServerStorage(connectionFactory);
+
+            Assert.NotNull(storage);
```
odinserj: We need to check that the given `connectionFactory` object is used to get a connection. You could use `Mock<DbConnection>` object to check it.

chinwobble:
One thing i Noticed is that in `SqlServerStorage.ToString()` will throw a null reference exception if you pass in a dedicated connection or connectionFactory

src/Hangfire.SqlServer/SqlServerStorage.cs
chinwobble: good suggestion! I've made this change

tests/Hangfire.SqlServer.Tests/SqlServerStorageFacts.cs
chinwobble: yep I've added this change. Let me know if it looks good to you now.

tests/Hangfire.SqlServer.Tests/SqlServerStorageFacts.cs
```diff
@@ -41,6 +43,28 @@ public void Ctor_CanCreateSqlServerStorage_WithExistingConnection()
...
+        }
+
+        [Fact, CleanDatabase]
+        public void CreateAndOpenConnection_UsesConnectionFactory()
+        {
+            var mockConnectionFactory = new Mock<Func<DbConnection>>();
+            var expectedConnection = ConnectionUtils.CreateConnection();
```
odinserj: You can create `Mock<DbConnection>` instead, pass `() => connection.Object` as a connection factory and check `CreateAndOpenConnection` returns the same `connection.Object` – currently this test anyway fails.

tests/Hangfire.SqlServer.Tests/SqlServerStorageFacts.cs
odinserj: Unfortunately this unit test doesn't test anything, other than *we don't have any exceptions* when creating an instance of the `SqlServerStorage` class – the result of a `new` operator can never be null anyway. 

You can safely remove this test, and optionally add another test that checks the constructor throws `ArgumentNullException`, when connection factory is null. And also another test, that throws the same exception, when options argument is null in factory-based constructor overload.

odinserj:
Also, Travis CI fails, because there were a lot of changes for the .NET Core 1.X build tools. So don't look at it before the `dev` branch is merged into the `master`.

tests/Hangfire.SqlServer.Tests/SqlServerStorageFacts.cs
chinwobble: thanks I got the test working now

tests/Hangfire.SqlServer.Tests/SqlServerStorageFacts.cs
chinwobble: done

odinserj:
Thanks for you changes, @chinwobble! I've modified unit test to also check that connection is the same, please see 4dfe0fc commit message. And also reverted commit with `ToString` changes, because it returned parse error for some reason, and we shouldn't establish a connection in the constructor anyway – it doesn't throw anyway in the current implementation, because the body of the `ToString` method is enclosed with a try/catch statement.

Issue #1324, opened 2019-01-09
Calls to ReadFormSafeAsync only returns values on first call


DC-jc:
In OwinRequestExtension there is a hack to 

> "prevent "Unable to cast object of type 'Microsoft.Owin.FormCollection'  to type 'Microsoft.Owin.IFormCollection'" exception, when internalized version does not match the current project's one."

The first time this method is called the Form collection is returned as it should without any problems. Subsequent calls to this method however returns an empty collection.

The root cause of this seems to be the removing of the FormCollection entry in context.Environment.
After the FormCollection entry is re-added to the collection, calls to ReadFormAsync again returns the correct values.

I'm speculating on the cause of the behavior here, but my guess is that ReadFormAsync is optimized to only parse the request body on the first call and adds the entry in the Environment collection. After that, ReadFormAsync just returns the already stored collection.

I haven't been able to reproduce the problem that the hack was supposed to fix so I'm don't really have any suggestions on any workarounds since any changes might break the original fix.

odinserj:
@DC-jc, could you tell me what problem is caused by the behavior you've described?

DC-jc:
@odinserj I ran in to problems when I tried to use the Hangfire.Core.Dashboard.Management extension. The same issue I had is described here: https://github.com/tracstarr/Hangfire.Core.Dashboard.Management/issues/2

The extension lets users set parameters and create new jobs, and the parameters are sent by POST to the server which then tries to retrieve them by calling `Task.Run(() => context.Request.GetFormValuesAsync(PARAMETERNAME)).Result.FirstOrDefault();` where context is DashboardContext. 

The problem was that the jobs always got created with the parameter value set to `null`.  

My current workaround is to add the Microsoft.Owin package to the affected extension and call 
```
var owinContext = new Microsoft.Owin.OwinContext(context.GetOwinEnvironment());    
var formValues = owinContext.Request.ReadFormAsync().Result;
```

odinserj:
Thanks for the details, now I understand this is definitely a problem. May be we should cache the `form` variable somewhere after reading it in https://github.com/HangfireIO/Hangfire/blob/053afff4b7fcfa2be060760b14879c5ac7a93204/src/Hangfire.Core/Dashboard/Owin/OwinDashboardRequest.cs#L45.

DC-jc:
That would solve my problem, and is essentially what the current Owin implementation does.
The current implementation looks like this:
```
[...]

Issue #386, opened 2015-06-05
Owin Cast Exception When Project Is Using Microsoft.Owin 3.0.1


ranstett:
I have a project that is using the UseWsFederationAuthentication to authenticate against Azure AD. It requires Microsoft.Owin 3.0.1. I verifiied the authentication was working, I then added Hangfire and the Dashboard to the project through Nuget. When I would go to the Dashbaord and try to execute something I would get the following error. 

System.InvalidCastException: Unable to cast object of type 'Microsoft.Owin.FormCollection' to type 'Microsoft.Owin.IFormCollection'.

After reviewing the stack trace it appeared it wasn't in the Hangfire code, so I downloaded the GitHub package and updated the Microsoft.Owin Nuget Package to the 3.0.1 (latest), I built the project and swapped out the DLLs, then everything started working.

tombrant:
Which version of Hangfire did you download and update?  Was it the latest 1.4.3?

ranstett:
I had installed 1.4.3 and this issue was present. I committed a version of 1.4.3 (https://github.com/ranstett/Hangfire/commits?author=ranstett) with the updated OWIN. I'm using it in a production environment now and haven't had any problems.

odinserj:
Hi guys, sorry for the loong-loong delay :( I've just released Hangfire 1.4.4 with fix for this issue. Hangfire 1.4.x still uses Microsoft.Owin 2.1.0 (otherwise previous user get this error), but don't use the cached form collection now, re-reading the form. Hangfire 1.5.0 will come with the new Microsoft.Owin, however I'll keep the hack to save users who use Microsoft.Owin 2.1.0.

odinserj:
There is a bug in 1.4.4, if you installed it today, please upgrade to 1.4.5 :frowning:

PR #1329, opened 2019-01-16
Fix to ensure Form values are properly returned

DC-jc:
Fix to ensure Form values are properly returned in case different versions of Microsoft.Owin gets sideloaded in the solution.

As discussed in https://github.com/HangfireIO/Hangfire/issues/1324 this commit ensures GetFormValuesAsync will return the correct form values, even if different versions of Microsoft.Owin gets sideloaded in a solution. Originally, sideloading Owin would cause a InvalidCastException if another middleware in the pipeline has made a call to ReadFormAsync. The previous workaround stopped working in later versions of Owin.


odinserj:
Thanks for your PR, @DC-jc, I've added some miscellaneous commits. Did you test the changes against the problem described in https://github.com/tracstarr/Hangfire.Core.Dashboard.Management/issues/2 and the original https://github.com/HangfireIO/Hangfire/issues/386 issue? We need to ensure first our change makes something useful, and don't introduce regressions. While theoretically it should work, practically it may not.

DC-jc:
Yes, I have tested this against both issues. Dashboard.Management passes arguments to invoked jobs correctly and sideloading Owin versions will return the correct values of IFormCollection.

odinserj:
Great, thanks a lot for testing! So I'm merging these changes and will release them with 1.6.22.

PR #1332, opened 2019-01-24
System.OutOfMemoryException at System.Type.GetType

drungrin:
That is a workaround for https://github.com/dotnet/coreclr/issues/20616. Since the data is serialized exactly, there is no need to be case insensitive.

### DIFF
--- a/src/Hangfire.Core/Storage/InvocationData.cs
+++ b/src/Hangfire.Core/Storage/InvocationData.cs
@@ -57,7 +57,7 @@ namespace Hangfire.Storage
         {
             try
             {
-                var type = System.Type.GetType(Type, throwOnError: true, ignoreCase: true);
+                var type = System.Type.GetType(Type, throwOnError: true, ignoreCase: false);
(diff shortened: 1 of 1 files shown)


odinserj:
Thanks for the hint, @drungrin, but I think we shouldn't add workarounds for other projects here, especially when these changes may break something (consider someone renamed a type in his or her assembly, and this change was working fine for some time). I've just released support for custom type resolvers with [Hangfire 1.6.22](https://www.hangfire.io/blog/2019/01/28/hangfire-1.6.22.html), so you can use a custom one:

```csharp
GlobalConfiguration.Configuration
    .UseXXX()
    .UseTypeResolver(typeName => Type.GetType(typeName, throwOnError: true, ignoreCase: false));
```

drungrin:
Closing the pull request since the external cause (dotnet/coreclr#20616.) has been closed.

odinserj:
Thanks for tracking this Michel!

PR #1334, opened 2019-01-24
Optimize implementation of recurring jobs

### LINKED ISSUE #167: Should be able to specify seconds in RecurringJobScheduler

### LINKED ISSUE #233: Trigger recurring task doesn't update the Last Executed Time
As the subject states, triggering a recurring task through the Dashboard doesn't update the Last Executed Time.

Is this as intended or a bug?

### LINKED ISSUE #391: dashboard recurring jobs sorting
the job list is not sorted on Id. i use alphanumeric Ids and they are randomly sorted.

### LINKED ISSUE #449: Performance problems with huge amount of recurring jobs (1.500.000 Jobs)
As posted before here (https://github.com/HangfireIO/Hangfire/issues/247), I have about 1.6 million recurring jobs.

I have the problem, that these recurring jobs, wont trigger in time, and I think I have found the couse for this.

It is coused by the Execute function of the RecurringJobScheduler (https://github.com/HangfireIO/Hangfire/blob/master/src/Hangfire.Core/Server/RecurringJobScheduler.cs#L57)

```
var recurringJobIds = connection.GetAllItemsFromSet("recurring-jobs");
```

scince I am using SQL Server as Database, the query looks like this :

```
select Value from HangFire.[Set] where [Key] = 'recurring-jobs'
```

This query fetches ALL (in my case 1.637.623) results from the database at once. So it takes some time to fetch all the results from the database. Even results, which are not necessary to fetch, are getting processed by the foreach.

I think about removing the LastExecution, NextExecution, as well as the TimeZoneId from the Hash table, and add them as a column to the Set table. So you can perform a query to not fetch jobs, that are already enqueued, and dont has to be enqueued.

### LINKED ISSUE #791: Recurring Job List: Default Sort Order
It appears the Recurring Job list on the Dashboard is ordered based on Date Created (or randomly/based on a hidden/internal unique id) which is fairly confusing / useless for end users.

Please can be it alphabetically ordered on the Id or Job columns or even better allow the column headings to be clicked to change the sorting?

### LINKED ISSUE #843: RecurringJobDto.NextExecution has delay in value getting set
It seems that the NextExecution value is only returned after the job has values in the Hash table. Consider parsing the cron schedule when there is no hash value yet and returning the next occurrence from there as a fallback. This would guarantee that there is always a NextExecution shown in the returned Dto.

### LINKED ISSUE #1147: when BackgroundJobServer dispose EveryMinuteThrottler throw OperationCanceledException
i use hangfire in console app,
when use ctrl + c to shotdown the app,  and after call BackgroundJobServer.Dispose(), 
then the OperationCanceledException threw by EveryMinuteThrottler

why?

### LINKED ISSUE #1262: Can you support timed task processing for seconds level units?
Is it possible to support seconds of time task processing, or does Hangfire support minutes or more of time task processing?

odinserj:
This PR contains a significant optimization for recurring jobs – we now using indexed fetch instead of iterating through all of the recurring jobs each minute and decide which to enqueue, and which don't. The implementation is almost fully similar to the `DelayedJobScheduler`'s now, and also contains improvements from PR #1330.

The implementation is fully backward compatible, and on the first run all the recurring jobs will be enqueued or scheduled to later time. After this kind of "migration", the pressure on a storage will be heavily reduced on next `RecurringJobScheduler` executions – it will fetch only those recurring jobs that will be scheduled.

And last but not least, this optimization allows us to use second-level cron expressions without hurting our database as much as previously. So if I find the easiest way to implement the feature, it will also come with 1.7.0.

Fixes #167,
Fixes #233.
Fixes #391.
Fixes #449.
Fixes #791.
Fixes #843.
Fixes #1147.
Fixes #1262.

**TODO**
* [x] Refactor duplicating code between scheduler and manager.
* [x] Add more unit tests for the `RecurringJobManager` class.
* [x] Add support for second-based cron expressions.


willchis:
Hey it looks like the `AddOrUpdate_AddsAJob_ToTheRecurringJobsSet` test is failing because a different `AddToSet()` overload is now being called.  Would something like the following work to check it's being called with the right parameters?
```
var now = DateTime.UtcNow;
var nextMinute = now.AddSeconds(-now.Second).AddMinutes(1);
_transaction.Verify(x => x.AddToSet(
    "recurring-jobs", 
    _id,
    JobHelper.ToTimestamp(nextMinute)));
```

Also the `AddOrUpdate_DoesNotUpdateCreatedAtValue_OfExistingJobs` test is failing because the job has no `CreatedAt` property in the test setup.  It passes if you give it this property in the setup:
```
 // Arrange
_connection.Setup(x => x.GetAllEntriesFromHash($"recurring-job:{_id}"))
    .Returns(new Dictionary<string, string>() { { "CreatedAt", JobHelper.SerializeDateTime(DateTime.UtcNow) } });
```

@odinserj I'd like to help but what's the etiquette with PRs like this? Is it useful if I fix or add new unit tests (as you mentioned in your TODO list)? Or would it be better to just look at contributing to another issue and to not interfere with this branch?!  Thanks and great project by the way!

odinserj:
Oh, nice catch @willchis, I didn't see that tests fail – build on AppVeyor should be tweaked a bit to fail in this case. To be honest, I don't know whether you can collaborate on this pull request, you can try to fork the `improve-recurring` branch and make PR on it, but I doubt it will change the current PR. 

However if you'd like to help, could you check why AppVeyor don't report build failures, when at least one test fails?

PR #1341, opened 2019-01-30
Adding check for each test run exit code.

willchis:
Creating pull request to test appveyor results in PR build/test run.


odinserj:
Thanks for your investigation on your issue, they helped me to realize that the best way is to avoid using pipelines in the build script and explicitly run all the test by their names, so here's the change – https://github.com/HangfireIO/Hangfire/commit/876e320007073e0a1fb36379a9f3bef445082590.

PR #1353, opened 2019-02-15
Fix IsBatchingAvailable() check

pieceofsummer:
and also cache result to not check it again on every iteration :)


odinserj:
Oh, I've just implemented this in 001ee0869d811ee34ec5f0af6d29ad2c47b26dc1 (recurring scheduler) and 3c2ec58805bd8f3e210047f667260f98df4fdcae (delayed scheduler) and 6dc93e83c514c03a6d4459af9ddd3c3c0c8c3e84 (add caching) commits. That implementation catches all non-expected exceptions (not only `NotSupportedException`) and provides caching per connection type implementation to avoid issues with multiple storages. 

Nevertheless thanks for reporting and for this PR, I'm planning to release the correction as 1.7.0-beta4 today.

PR #1354, opened 2019-02-15
Implement IsBatchingAvailable check as a simple property

pieceofsummer:
This way we don't need expensive runtime checks and duplicate logic in `RecurringJobScheduler` / `DelayedJobScheduler`, and those storages that actually implement batching may just override it in their `JobStorage` subclasses. It's not as if any storages (except for SqlServer, and maybe Redis from HF Pro) support it right now. But it would be much harder to change it later, when this feature would gradually become adopted by 3rd-party storages.

I'm not quite sure if this should be a property of `JobStorage` though. It might seem reasonable to make it a property of `JobStorageConnection` instead, but that would require safe cast before we can actually check it.

### DIFF
--- a/src/Hangfire.Core/JobStorage.cs
+++ b/src/Hangfire.Core/JobStorage.cs
@@ -71,6 +71,8 @@ namespace Hangfire
             }
         }
 
+        public virtual bool IsBatchingAvailable => false;
+
         public abstract IMonitoringApi GetMonitoringApi();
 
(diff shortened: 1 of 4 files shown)


odinserj:
`IsBatchingAvailable` is a bad name in the context of the `JobStorage` class. It have more or less sense in schedulers, because its meaning have more narrow scope there.

PR #1360, opened 2019-02-22
Use serilog SourceContext naming convention

### LINKED ISSUE #1359: Serilog context property name does not follow Serilog convention.
In current source, at time of writing:
https://github.com/HangfireIO/Hangfire/blob/23a71c049b298a144a21bb6460f3d18eb4708bc6/src/Hangfire.Core/App_Packages/LibLog.1.4/LibLog.cs#L1043

This piece of codes registers a Serilog log event property with the name `"Name"`. Serilog convention is to call this property `"SourceContext"`. This is important because the Serilog MinimumLevel.Override function specifically targets the `SourceContext` property.

Example serilog configuration code:

```csharp
Log.Logger = new LoggerConfiguration()
    .MinimumLevel.Debug()
    .MinimumLevel.Override("Hangfire", LogEventLevel.Error) // this doesn't work
    .MinimumLevel.Override("Microsoft", LogEventLevel.Error) // this works
```

The only alternative to `MinimumLevel.Override` is to filter out any message that has a `Name` property with a value starting with "Hangfire", but that removes all log messages regardless of log level, meaning even error and fatal messages are removed.

Possible solutions:

1.  Update LibLog version 1.4 to a newer version. I have checked the latest version, 5.0.6 and that correctly uses `"SourceContext"`.
2. Change the name in your local copy of LibLog. This has less impact but would fix this issue.

pobiega:
This fixes #1359.

### DIFF
--- a/src/Hangfire.Core/App_Packages/LibLog.1.4/LibLog.cs
+++ b/src/Hangfire.Core/App_Packages/LibLog.1.4/LibLog.cs
@@ -1040,7 +1040,7 @@ namespace Hangfire.Logging.LogProviders
                 valueParam,
                 destructureObjectsParam
             }).Compile();
-            return name => func("Name", name, false);
+            return name => func("SourceContext", name, false);
         }
 
         internal class SerilogCallbacks


odinserj:
Thanks @pobiega! This fix will be released in 1.6.23.

PR #1374, opened 2019-03-19
Replacing ContainsKey with TryGetValue to avoid duplicate lookups in …

robink-teleopti:
…dictionaries


robink-teleopti:
Yes, actually, I didn't mention it, but I think there was a minor bug in the RecurringJobEntity code before my change; `TimeZone = recurringJob.ContainsKey("TimeZoneId") && !String.IsNullOrWhiteSpace("TimeZoneId")` where it checked the property name for null or whitespace. Might be good to know though.

Issue #1069, opened 2017-11-30
ExpirationManager - System.ArgumentOutOfRangeException - Parameter name: timeout


mberacochea:
Hi,

Hangfire is throwing an error constantly and I can't figure out what is happening. The app is working and Jobs are executed as expected.

I'm using:

- Hangfire version=1.6.17 targetFramework=net45
- Hangfire.Core version=1.6.17 targetFramework=net45
- Hangfire.Dashboard.Authorization version=2.1.0 targetFramework=net45
- HangFire.Ninject version=1.2.0 targetFramework=net45
- Hangfire.SqlServer version=1.6.17 targetFramework=net45

My config:

```C#
Hangfire.GlobalConfiguration.Configuration.UseNinjectActivator(CreateKernel());

Hangfire.GlobalConfiguration.Configuration.UseSqlServerStorage("HangfireConnection", new Hangfire.SqlServer.SqlServerStorageOptions
{
	JobExpirationCheckInterval = TimeSpan.FromDays(30),
});

app.UseHangfireDashboard("/hangfire");
app.UseHangfireServer();
```

On the log I'm getting:
```C#
2017-11-30 18:06:53,983 ERROR Hangfire.SqlServer.ExpirationManager - Error occurred during execution of 'Hangfire.SqlServer.ExpirationManager' process. Execution will be retried (attempt #14) in 00:03:14 seconds.
System.ArgumentOutOfRangeException: Number must be either non-negative and less than or equal to Int32.MaxValue or -1.
Parameter name: timeout
   at System.Threading.WaitHandle.WaitOne(TimeSpan timeout, Boolean exitContext)
   at System.Threading.WaitHandle.WaitOne(TimeSpan timeout)
   at Hangfire.SqlServer.ExpirationManager.Execute(CancellationToken cancellationToken)
   at Hangfire.Server.AutomaticRetryProcess.Execute(BackgroundProcessContext context)
```

pieceofsummer:
30 days is too much for `JobExpirationCheckInterval`. It would be `2 592 000 000` milliseconds (30 * 24 * 60 * 60 * 1000), which is larger than `Int32.MaxValue`.

`JobExpirationCheckInterval` determines how often Hangfire will check if there are some jobs to clean up, and it shouldn't typically exceed 1 hour.

You've probably mistaken it for `JobExpirationTimeout`, which determines how long the jobs live after completion. That could be set with an `IElectStateFilter`.

[...]

PR #1377, opened 2019-03-22
Add verification to job expiration check interval

carlowahlstedt:
If this value is over int.MaxValue then continual errors will be logged as reported in #1069. This will tell the consumer upfront that they have an issue with their settings. Verified based on the [docs for WaitOne](https://docs.microsoft.com/en-us/dotnet/api/system.threading.waithandle.waitone?view=netframework-4.7.2#System_Threading_WaitHandle_WaitOne_System_TimeSpan_).

```
Error occurred during execution of 'Hangfire.SqlServer.ExpirationManager' process. Execution will be retried (attempt #77) in 00:05:00 seconds.

System.ArgumentOutOfRangeException: Number must be either non-negative and less than or equal to Int32.MaxValue or -1.Parameter name: timeout   
at System.Threading.WaitHandle.WaitOne(TimeSpan timeout, Boolean exitContext)   at System.Threading.WaitHandle.WaitOne(TimeSpan timeout)   
at Hangfire.Common.CancellationTokenExtentions.Wait(CancellationToken cancellationToken, TimeSpan timeout)   
at Hangfire.SqlServer.ExpirationManager.Execute(CancellationToken cancellationToken)   
at Hangfire.Server.AutomaticRetryProcess.Execute(BackgroundProcessContext context)

<properties><property key='SourceContext'>Hangfire.SqlServer.ExpirationManager</property></properties>
```

### DIFF
--- a/src/Hangfire.SqlServer/SqlServerStorageOptions.cs
+++ b/src/Hangfire.SqlServer/SqlServerStorageOptions.cs
@@ -86,7 +86,18 @@ namespace Hangfire.SqlServer
 
         public bool PrepareSchemaIfNecessary { get; set; }
 
-        public TimeSpan JobExpirationCheckInterval { get; set; }
+        public TimeSpan JobExpirationCheckInterval 
+        {
+            get { return _jobExpirationCheckInterval; } 
+            set {
+                if (value.TotalMilliseconds > int.MaxValue)
+                {
+                    throw new ArgumentOutOfRangeException("Job expiration check interval cannot be greater than int.MaxValue");
+                }
+                _jobExpirationCheckInterval = value;
+            }
+        }
+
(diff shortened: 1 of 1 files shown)


carlowahlstedt:
This seems to be failing for some reason other than the changes here.

odinserj:
Thanks for your work on this, @carlowahlstedt! I've just merged PR, and the fix will be released in 1.6.24 tomorrow.

carlowahlstedt:
Just verified 
image

pieceofsummer:
Shouldn’t we have lower bound checked here as well? At the very least the timeout value should be positive.

Infinite timeout (-1) would just hang the ExpirationManager after the first iteration.

Zero values should be rejected as well, so ExpirationManager doesn’t run non-stop.

PR #1393, opened 2019-04-08
Concept for readonly with authorization

carlowahlstedt:
In implementing readonly for a dashboard I found it was an all or nothing deal. So you could be authorized but it was still readonly for you. Seems readonly would be a concept for anyone going to the dashboard and authorization would be a concept for editing the dashboard?

So I just created a conceptual implementation of making it readonly if it's marked that way AND if you are not authorized.

### DIFF
--- a/src/Hangfire.Core/Dashboard/DashboardContext.cs
+++ b/src/Hangfire.Core/Dashboard/DashboardContext.cs
@@ -15,6 +15,7 @@
 // License along with Hangfire. If not, see <http://www.gnu.org/licenses/>.
 
 using System;
+using System.Linq;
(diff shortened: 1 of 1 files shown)


pieceofsummer:
`IsReadOnly` is supposed to distinguish between administrators (who can delete/cancel/re-schedule jobs) and e.g. help desk staff who just monitor if everything is working. Even a read-only dashboard may contain sensitive info, so it was never intended to be exposed to any unauthorized client.

If you really want to expose it like that, you may simply provide an empty authorization filters list, and move the actual authorization logic to `IsReadOnlyFunc`.

PR #1420, opened 2019-05-03
Passing params in job creation phase

brian-knoll-micronetonline:
Updated PR of https://github.com/HangfireIO/Hangfire/pull/570


odinserj:
Thanks @brian-knoll-micronetonline. Unfortunately it contains breaking changes, so I'm unable to land it before Hangfire 2.0.

brian-knoll-micronetonline:
@odinserj Thanks for the reply. So will it get accepted after hangfire 2.0? That would be great if it does. If not is there any change i should make to get it accepted?

mjl321:
Is this still something being considered for 2.0? Would be useful functionality to have, it's not listed here though: https://github.com/HangfireIO/Hangfire/milestone/17

brian-knoll-micronetonline:
@mjl321 equivalent functionality was added in https://github.com/HangfireIO/Hangfire/releases/tag/v1.8.0

mjl321:
Thanks for the quick reply. I'm struggling to find any documentation / code for this functionality. I'm adding a recurring job, but can't see any overloads for RecurringJob.AddOrUpdate that allow the setting of Job parameters, I can't spot anything in the BackgroundJobClient either. It looks like there's a new parameter defined in IBackgroundJobClientV2. I'd be grateful if you could point me in the right direction for adding job parameters when creating a recurring job.

mjl321:
Awesome, thank you. I've changed it so the recurring job now calls an intermediary method which then uses the new(?) IBackgroundJobClientV2 to create the job with the parameter (tenantid). 

So I've now got a Recurring Job which calls an AddJobForTenant method with the tenant info and a class that provides the actual method to call (as couldn't pass delegate as job method parameter), which creates the actual job (with the tenant id as a job parameter which is later used to set the tenant context when the job is run).

PR #1424, opened 2019-05-06
fix minor bug in DefaultInstall.sql

penenkel:
TSQL is strange and therefor `CASE @CURRENT_SCHEMA_VERSION WHEN NULL THEN 'none' ELSE CONVERT(nvarchar, @CURRENT_SCHEMA_VERSION) END;` does not work as one might expect. To correctly handle NULL we must use `CASE WHEN @CURRENT_SCHEMA_VERSION IS NULL THEN 'none' ELSE CONVERT(nvarchar, @CURRENT_SCHEMA_VERSION) END;` instead.


odinserj:
Thanks @penenkel, definitely. Could you also modify the `Install.sql` file to include this change?

odinserj:
I've modified the `DefaultInstall.sql` file as well. Will be shipped in 1.7.3.

PR #1429, opened 2019-05-08
Added Dutch language, and updated missing translation on server page

r-win:


odinserj:
Thanks @r-win for taking care of it! I'm merging this change, and it will be shipped in 1.7.3 by the end of May.

PR #1431, opened 2019-05-10
Added Cron.Never()

michaltalaga:
conveniance method for creating a never firing recuring job.


michaltalaga:
need to merge into dev

PR #1432, opened 2019-05-10
Added Cron.Never()

michaltalaga:
convenience method for never-firing recurring jobs

### DIFF
--- a/src/Hangfire.Core/Cron.cs
+++ b/src/Hangfire.Core/Cron.cs
@@ -214,6 +214,15 @@ namespace Hangfire
             return $"{minute} {hour} {day} {month} *";
         }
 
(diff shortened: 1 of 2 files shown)


odinserj:
Thanks @michaltalaga, but should it be added as a `Cron.Never` or it's better to allow to pass `null` argument for the `cron` parameter instead?

michaltalaga:
Both have pros and cons. null is more of a "need to know and read the docs" vs more explicit Cron.Never()

Also Cron.Never() could be reused in other places if needed.

odinserj:
Thanks, @michaltalaga. I'm merging this to `master` branch as the new method has minimal impact anyway. Will be released in 1.7.3.

Issue #550, opened 2016-04-14
Queue name validation


pzielinski86:
We added a task with a queue name containing a dash ('-'). We then tried to trigger it on the dashboard but it just hang on _"Triggering..."_ and it did nothing.

After investigating, we found that [there is a regular expression](https://github.com/HangfireIO/Hangfire/blob/2126550aaf52870122548474b89b1630dbb7d2c8/src/Hangfire.Core/States/EnqueuedState.cs#L226), permitting only letters, digits and underscore.

1. Is this is a bug that it accepts a task with a dash (in the queue name) but can't trigger it?
2. If it's not a bug, then maybe there should be an error on the dashboard explaining why.

Thanks

pzielinski86:
We have just discovered that it doesn't even accept upper case. For example "Test" is invalid name (it should be "test").

AbubakerB:
Yes, i have the same issue!
It seems odd that it accepts it but then can't run it

IanYates:
See the docs: http://docs.hangfire.io/en/latest/background-processing/configuring-queues.html

```
The Queue name argument must consist of lowercase letters, digits and underscore characters only.
```

So I guess, at the least, an exception should be thrown if you supply a name that doesn't conform.

odinserj:
Closing – was fixed by PR #561 in 1.5.7.

PR #1469, opened 2019-07-21
Add support for dashes `-` in queue names

caioproiete:
As the title says...

Relates to #550 and #561


odinserj:
Thanks @caioproiete, especially for making corresponding changes in documentation!

Issue #1487, opened 2019-08-09
net461 TargetFramework for Hangfire.AspNetCore


danstur:
Would it be possible to add net461 as a target framework for the Hangfire.AspNetCore csproj? The problem is that people using .NET 4.6.1 or higher get the Net451 nuget which excludes many useful helper functions such as `AddHangfireServer`, since these are hidden behind a `#IF NETSTANDARD2_0` check. 

.NET 4.6.1 and higher implement NetStandard 2.0, so the same code could be used. Adding net461 to the TargetFrameworks for the project and expanding the #IF (preferably via an additional property?) checks would allow people to use all this functionality. 

Currently we're simply copy-pasting the parts we need into our own project, but clearly this is a suboptimal solution. I'd be happy to provide a PR that handles this.

odinserj:
Sure, will be happy to accept that PR! One side effect here is that project that target .NET 4.6.1 and still sitting on ASP.NET Core 1.X will be unable to use the newer version. However I don't see any reason to still use those early versions – there were too many bugs and non-implemented features.

danstur:
@odinserj I guess with the right app.config redirections they could get it to work, but yeah I doubt there's much interest in that configuration in any case. I'll try to get the PR done ASAP - looks pretty simple on first glance :-)

PR #1488, opened 2019-08-09
Enable Net461 for Hangfire.AspNetCore

danstur:
Adds net461 as a target framework and creates a new NuGet configuration for it. 

See #1487 for rationale.

### DIFF
--- a/psake-project.ps1
+++ b/psake-project.ps1
@@ -51,6 +51,8 @@ Task Collect -Depends Test -Description "Copy all artifacts to the build folder.
     Collect-Assembly "Hangfire.SqlServer" "netstandard2.0"
     Collect-Assembly "Hangfire.AspNetCore" "netstandard2.0"
     Collect-Assembly "Hangfire.NetCore" "netstandard2.0"
+	
+	Collect-Assembly "Hangfire.AspNetCore" "net461"
     
     Collect-Content "content\readme.txt"
     Collect-Tool "src\Hangfire.SqlServer\DefaultInstall.sql"
--- a/psake-project.ps1
+++ b/psake-project.ps1
@@ -51,8 +51,8 @@ Task Collect -Depends Test -Description "Copy all artifacts to the build folder.
     Collect-Assembly "Hangfire.SqlServer" "netstandard2.0"
     Collect-Assembly "Hangfire.AspNetCore" "netstandard2.0"
     Collect-Assembly "Hangfire.NetCore" "netstandard2.0"
-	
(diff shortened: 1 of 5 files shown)


danstur:
I have little experience using psake for build automation, so I basically just looked at the general structure and adapted it for net461 - looks reasonable, but if there's something I missed I'd be happy for guidance in that regard.

psake-project.ps1
```diff
@@ -51,6 +51,8 @@ Task Collect -Depends Test -Description "Copy all artifacts to the build folder.
     Collect-Assembly "Hangfire.SqlServer" "netstandard2.0"
     Collect-Assembly "Hangfire.AspNetCore" "netstandard2.0"
     Collect-Assembly "Hangfire.NetCore" "netstandard2.0"
+	
+	Collect-Assembly "Hangfire.AspNetCore" "net461"
```
odinserj: Please replace tab with four spaces 😊

src/Hangfire.AspNetCore/BackgroundJobServerHostedService.cs
```diff
@@ -1,4 +1,4 @@
-﻿#if NETSTANDARD2_0
+﻿#if ASPNETCORE2_AVAILABLE
```
odinserj: Some of AspNetCore files are also linked with the `Hangfire.NetCore` project. I see the `ASPNETCORE2_AVAILABLE` is introduced only in the `Hangfire.AspNetCore` project, and not in the `Hangfire.NetCore`, and some of the classes and methods in the latter project will be missing after this change. And this should be fixed. 

I think the `NETSTANDARD2_0 || NET461` will be enough.

psake-project.ps1
danstur: Mea culpa, fixed.

src/Hangfire.AspNetCore/BackgroundJobServerHostedService.cs
danstur: Missed that, I wanted to avoid duplicating the logic in multiple places, but if the files are shared this gets cumbersome too. Fixed.

odinserj:
Thank you @danstur, looks brilliantly! I'm merging the changes, they will be shipped in version 1.7.7 this month.

cuzzlor:
Hello @odinserj, 

Any update on the release of 1.7.7?

I've just been trying to migrate an app targeting net461 and found I could not use `services.AddHangfireServer()`.

Thank you!

PR #1489, opened 2019-08-13
Return the BackgroundJobId when triggering a RecurringJob

omeaart:
W're triggering recurring jobs programmatically and would like to use the created jobid after that. Otherwise we have to query that by hand from the database.


burningice2866:
You can't go around changing public interfaces like this. There might be 3rd party vendors doing their own implementations that will break. There is also the issue about binary compatibility when you change the signature of a method.

PR #1506, opened 2019-09-11
I found that the most resource hogger for our sql server is the sp_ge…

safigi:
…tapplock. This server is a huge report server with many data load (importing data from the cloud), and we get the the most wait time is sp_getapplock.

I would recommend a new option and we can dequeue jobs without applock. WITH (ROWLOCK) for update should be enough


odinserj:
Blocking fetch is only used when `SqlServerStorageOptions.QueuePollInterval` value is less than one second. You can simply use any value starting from `TimeSpan.FromSeconds(1)` to avoid calls to the `sp_getapplock` procedure.

PR #1509, opened 2019-09-13
Improve display of job result and properties

0xced:
It's quite common that job properties and/or result are strings. Since everything is serialized as JSON in the database, a string becomes quoted and has escaped characters. For displaying in the dashboard, it's nicer without the quotes and escaped characters.

Before:
Before

After:
After

### DIFF
--- a/src/Hangfire.Core/Dashboard/HtmlHelper.cs
+++ b/src/Hangfire.Core/Dashboard/HtmlHelper.cs
@@ -350,5 +350,17 @@ namespace Hangfire.Dashboard
         {
             return WebUtility.HtmlEncode(text);
         }
+
+        public string HtmlEncodePretty(string text)
+        {
+            try
+            {
+                return HtmlEncode(SerializationHelper.Deserialize<string>(text, SerializationOption.User)).Replace("\n", "<br/>");
+            }
+            catch (Exception)
+            {
+                return HtmlEncode(text);
+            }
+        }
     }
 }
--- a/src/Hangfire.Core/Dashboard/JobHistoryRenderer.cs
+++ b/src/Hangfire.Core/Dashboard/JobHistoryRenderer.cs
@@ -196,7 +196,7 @@ namespace Hangfire.Dashboard
             if (stateData.ContainsKey("Result") && !String.IsNullOrWhiteSpace(stateData["Result"]))
             {
                 var result = stateData["Result"];
-                builder.Append($"<dt>Result:</dt><dd>{html.HtmlEncode(result)}</dd>");
+                builder.Append($"<dt>Result:</dt><dd>{html.HtmlEncodePretty(result)}</dd>");
 
                 itemsAdded = true;
             }
--- a/src/Hangfire.Core/Dashboard/Pages/JobDetailsPage.cshtml
+++ b/src/Hangfire.Core/Dashboard/Pages/JobDetailsPage.cshtml
@@ -1,4 +1,4 @@
-﻿@* Generator: Template TypeVisibility: Internal GeneratePrettyNames: True *@
+@* Generator: Template TypeVisibility: Internal GeneratePrettyNames: True *@
 @using System
 @using System.Collections.Generic
 @using System.Linq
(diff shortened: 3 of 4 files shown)


0xced:
Ping.

src/Hangfire.Core/Dashboard/HtmlHelper.cs
```diff
@@ -326,5 +326,17 @@ public string HtmlEncode(string text)
         {
             return WebUtility.HtmlEncode(text);
         }
+
+        public string HtmlEncodePretty(string text)
+        {
+            try
+            {
+                return HtmlEncode(SerializationHelper.Deserialize<string>(text)).Replace("\n", "<br/>");
```
pieceofsummer: `SetJobParameter` and `Result` use `SerializationOption.User` option for serialization, so it should be also used to deserialize them.

0xced: Good catch, I addressed this issue in dfe1dd670c60b2fd6ddb9398bed91a95e9ac4485.

0xced:
@odinserj Do you have any plan to merge this?

0xced:
Ping @odinserj.

0xced:
Freshly rebased onto master in case there's interest in merging this pull request.

0xced:
I just rebased on master. Is there a chance to see this merged @odinserj or @pieceofsummer?

0xced:
Rebased again on master. Still crossing my fingers that this pull request might get some attention. 🤞

0xced:
Freshly rebased on the `dev` branch, still crossing my fingers that this might get some attention. Ping @odinserj.

The failing test on AppVeyor (Ubuntu only, passes on Windows) seems completely unrelated to this pull request:
```
Failed Hangfire.SqlServer.Tests.SqlServerTimeoutJobFacts.Timer_UpdatesFetchedAtColumn(useMicrosoftDataSqlClient: False) [10 s]
  Error Message:
   Assert.Equal() Failure
Expected: 2022-02-01T13:31:25.3700000
Actual:   2022-02-01T13:31:27.3700000
```

odinserj:
Thanks @0xced, I will check this before releasing 1.8 next month!

odinserj:
Hello @0xced! 

With Hangfire [1.8.0-rc3](https://www.hangfire.io/blog/2023/03/30/hangfire-1.8.0-rc3.html) it is possible to register custom widgets on the Job Details page, so the feature requested in this PR can be implemented in the following way:

```csharp
.UseJobDetailsRenderer(10, dto =>
{
    var succeededState = dto.JobDetails.History
        .FirstOrDefault(x => x.StateName == SucceededState.StateName);

    if (succeededState != null && succeededState.Data.TryGetValue("Result", out var result))
    {
        var builder = new StringBuilder();
        builder.Append("<h3>Last Result</h3>");
        builder.Append("<pre><code>");
        builder.Append(WebUtility.HtmlEncode(SerializationHelper
            .Deserialize<string>(result, SerializationOption.User)
            .Replace("\n", "<br/>")));
        builder.Append("</code></pre>");
        return new NonEscapedString(builder.ToString());
    }

    return null;
})
```

This gives the following results without changing anything in Hangfire itself and allows more features to be implemented:

image

I think we can close this PR now, what do you think?

0xced:
Thank you, it works fine for rendering the state history. I don't actually need to add a widget, I just want to remove the useless quotes on the `Result`. I was able to achieve this goal with a custom renderer that only alters the history:

```cs
.UseJobDetailsRenderer(10, dto =>
{
    foreach (var history in dto.JobDetails.History.Where(e => e.StateName == SucceededState.StateName))
    {
        if (history.Data.TryGetValue("Result", out var result))
        {
            history.Data["Result"] = result.Trim('"');
        }
    }

    return null;
});
```

Unfortunately, this callback is called _after_ the parameters are rendered so modifying `dto.JobDetails.Properties` has no effect. We can't trim or otherwise modify how the parameters are rendered by altering the  `JobDetailsRendererDto`. 😞

Is there an alternative solution to tweak the job parameters?

PR #1527, opened 2019-10-01
Add GetInstallScript method

altso:
This PR exposes `GetInstallScript()` method on `SqlServerObjectInstaller`, so it's easy to get programmatic access to prepared installation script. It could be used in EF migrations like this:
```cs
public partial class Hangfire : Migration
{
    protected override void Up(MigrationBuilder migrationBuilder)
    {
        migrationBuilder.Sql(SqlServerObjectsInstaller.GetInstallScript("dbo", false));
    }

    protected override void Down(MigrationBuilder migrationBuilder)
    {
        // not supported. remove manually
    }
}
```


altso:
Thank you!

PR #1537, opened 2019-10-12
Recurring Jobs page added start and stop button

yunuskiran:
Recurring Jobs page added start and stop button with actions. IsActive key added to hash table and when enque check to IsActive key. When is false job doesn't execute codes else job execute.

hangfire-pull-request


yunuskiran:
.

PR #1539, opened 2019-10-15
Create i18n to js files

Alexandre-Machado:
image


samples/ConsoleSample/Program.cs
```diff
@@ -17,7 +17,7 @@ public static void Main()
                 .SetDataCompatibilityLevel(CompatibilityLevel.Version_170)
                 .UseSimpleAssemblyNameTypeSerializer()
                 .UseRecommendedSerializerSettings()
-                .UseSqlServerStorage(@"Server=.\;Database=Hangfire.Sample;Trusted_Connection=True;", new SqlServerStorageOptions
+                .UseSqlServerStorage(@"Server=(LocalDb)\MSSQLLocalDB;Database=Hangfire.Sample;Trusted_Connection=True;", new SqlServerStorageOptions
```
Alexandre-Machado: this change is to use the embedded Sql Server on Visual Studio

Alexandre-Machado:
do you think this is a good job @odinserj?

odinserj:
Thank you for your work @Alexandre-Machado! This is a perfectly valid scenario, but I doubt we should introduce a new way to handle internationalization. I think we should find the simplest way to expose localized strings via embedded `*.resx` files and load localized strings into JavaScript somehow (for example, by generating a `*.js` file with required strings).

burningice2866:
My common approach is to create a simple json-generating handler which returns structure resembling the resx-file in use within a given language context. An everywhere in javascript where you need to show some text, you refer to this json-structure using the same resx-keys as you would have on the server.

odinserj:
Thanks Pauli, something like this, but I'm afraid JSON will require yet another HTTP GET invocation, and with plain JavaScript file we can automate the behavior. Also there's no need to expose all the resources in that JSON or JavaScript file, only the required ones.

Alexandre-Machado:
I have some points to stand up for my way:
 - in the same `resx`, we have a lot of unused strings
 - in another `resx`, even so, we have a new build process
 - in `resx`, we don't have a template string (at least not in an elegant way)

Alexandre-Machado:
It is acceptable to me that the backend application has a different way of dealing with its internationalization. even because plugins like `moment` and others already use` * .js` files for this purpose.
It sounds like a good move for us to have more innovation on the hangfire frontend.

samples/ConsoleSample/Program.cs
Alexandre-Machado:
```suggestion
                .UseSqlServerStorage(@"Server=.\;Database=Hangfire.Sample;Trusted_Connection=True;", new SqlServerStorageOptions
```

Alexandre-Machado:
@odinserj, let me know if you still consider this PR or close it, please

odinserj:
> in the same *.resx, we have a lot of unused strings

We can generate a custom `*.js` file with the subset of strings that known to be required in our UI.

> It is acceptable to me that the backend application has a different way of dealing with its internationalization. even because plugins like moment and others already use *.js files for this purpose.

Localization for those plugins is maintained outside of this project, and we can't change the way how they are localized (and don't need as externally maintained).

I will try to implement a simpler solution after a while, and if there will be more downsides, I will merge this PR. I appreciate your time and your work, but don't like to complicate things in the long term.

Alexandre-Machado:
Thanks for the appreciation and sorry for any inconvenience.

I already have worked with a `.resx` file build into a `js` format and am not convinced that this is the best strategy for dealing with the` js` application layer. because I think it would be interesting to have a stronger separation between the `c#` application and the `js` application.

PR #1231, opened 2018-07-18
Allow modification of the dashboard title

tbertenshaw:
We have several dashboards all exposed using ports and without meaningful urls. Allowing modification of the title on the dashboard to a more meaningful title than "Hangfire Dashboard" will make it easier to discover the purpose.

PR #1545, opened 2019-10-18
Add HTML tags on Dashboard title

caioproiete:
Small improvement upon #1231 to allow HTML tags on the title of the dashboard.


odinserj:
Thanks @caioproiete, merging. Looks like it's for referencing other dashboards, right?

PR #1557, opened 2019-11-05
Adding prefix to table names

girakun:
We have some old running software connecting to the same SQL database as our new software. The old software do some stuff that ignores schema and as some of our tables has the same name as some of Hangfires it cause problems. I did also see two old pull request where someone where looking for something similar to this. (Pull request 1097 and 1301)

We have not upgraded Hangfire for a while, but we have used this since we started using Hangfire two years ago. I now updated our code base from the latest version and i figured i could share this, if this is interesting for someone else. 

It is done in the quickest way possible (search and replace) and is probably not the best implementation, but it has been working for us. It's possible to create a better implementation (Example: Something that doesn't require manual add of the prefix before the table every time) but then i want to be sure that it is a feature that will be taken in consideration for the official version, before i put the time in to do so. 

Quick about the implementation that is made so far. It's implemented in the same way custom scheme name is implemented. It has support for changing prefix over time as of now it is running when PrepareSchemaIfNecessary = true, maybe this should be moved to some other place or only run when EnableHeavyMigrations = true. Default there is no prefix. If prefix is set and tables do not correspond with prefix it will cause and exception at initialization of SQL storage.

### DIFF
--- a/src/Hangfire.SqlServer/Constants.cs
+++ b/src/Hangfire.SqlServer/Constants.cs
@@ -3,5 +3,6 @@
     internal class Constants
     {
         public static readonly string DefaultSchema = "HangFire";
-    }
+        public static readonly string DefaultTablePrefix = string.Empty;
+	}
 }
--- a/src/Hangfire.SqlServer/CountersAggregator.cs
+++ b/src/Hangfire.SqlServer/CountersAggregator.cs
@@ -100,11 +100,11 @@ BEGIN TRAN
 
 DELETE TOP (@count) C
 OUTPUT DELETED.[Key], DELETED.[Value], DELETED.[ExpireAt] INTO @RecordsToAggregate
-FROM [{storage.SchemaName}].[Counter] C WITH (READPAST, XLOCK, INDEX(0))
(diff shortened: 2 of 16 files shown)


pieceofsummer:
Is there any reason you cannot use a separate database for Hangfire storage, so table names don't overlap with your software?

CzechsMix:
@pieceofsummer 

I had a similar question, and I *could* use a separate database for hangfire storage, but I have multiple services each with their own database.

I don't like the idea of using a single hangfire storage for all services, but it seems like a large effort to have a hangfire datbase for each service that will use hangfire.

dbo.ServiceA
dbo.ServiaA_Hangfire
dbo.ServiceB
dbo.ServiceB_Hangfire

Ideally, I could just point hangfire at the SqlServer database for that application, and it would just prefix it's tables with "Hangfire_" or something to not conflict with the other tables.

If a second database for each service that uses hangfire is the way to go then I guess I could make due, but ideally I'd like to specify a prefix.

burningice2866:
Use Schemas!

PR #1575, opened 2019-11-29
add netcoreapp3.0 target for .Net Core 3 compatibility

stebueh:
When using Hangfire from an ASP.NET Core 3.0 application, I encountered a version conflict for the 'Microsoft.Extensions.Logging.Abstractions' package. 

Please consider this PR with the following changes for .Net Core 3 compatibility:

- add netcoreapp3.0 target
- add framework reference to 'Microsoft.AspNetCore.App' 
- use 'IHostApplicationLifetime' instead of deprecated 'IApplicationLifetime'

### DIFF
--- a/src/Hangfire.AspNetCore/Hangfire.AspNetCore.csproj
+++ b/src/Hangfire.AspNetCore/Hangfire.AspNetCore.csproj
@@ -1,6 +1,6 @@
 ﻿<Project Sdk="Microsoft.NET.Sdk">
   <PropertyGroup>
-    <TargetFrameworks>net451;net461;netstandard1.3;netstandard2.0;</TargetFrameworks>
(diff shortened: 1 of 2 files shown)


odinserj:
@stebueh thank you for your work. Could you tell me more about the conflict related to the Microsoft.Extensions.Logging.Abstractions package? Did you have a compile-time or run-time error? Was you able to resolve it without modifying Hangfire?

odinserj:
Now I see that for ASP.NET Core 3.0-related packages it's fine to reference `netcoreapp3.0` target from libraries. So since some interfaces were deprecated, it's much safer to use the newer version. I will merge this PR, but will try to remove the `Microsoft.AspNetCore.App` reference anyway, because it's too heavy. Thank you for this PR!

PR #1585, opened 2019-12-18
Added null checks for expired failed jobs

ppkale1995-cimpress:
We're using hangfire with redis in our production. To resolve the OOM exception we've added a filter to expire failed jobs. The dashboard on "hangfire/jobs/failed" gives NullReferenceException when the jobs get expired and are deleted. I saw that the null checks for job.Value are not added for FailedJobs because hangfire doesn't expect failed jobs to be expired.

### DIFF
--- a/src/Hangfire.Core/Dashboard/Pages/FailedJobsPage.cshtml
+++ b/src/Hangfire.Core/Dashboard/Pages/FailedJobsPage.cshtml
@@ -73,8 +73,8 @@
                                     </th>
                                 }
                                 <th class="min-width">@Strings.Common_Id</th>
-                                <th>@Strings.FailedJobsPage_Table_Failed</th>
                                 <th>@Strings.Common_Job</th>
+                                <th class="align-right">@Strings.FailedJobsPage_Table_Failed</th>
                             </tr>
                         </thead>
                         <tbody>
@@ -106,12 +106,6 @@
                                 }
                                 else
                                 {
-                                    <td class="min-width">
-                                        @if (job.Value.FailedAt.HasValue)
-                                        {
-                                            @Html.RelativeTime(job.Value.FailedAt.Value)
-                                        }
-                                    </td>
                                     <td>
                                         <div class="word-break">
                                             @Html.JobNameLink(job.Key, job.Value.Job)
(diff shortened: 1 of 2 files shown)


odinserj:
Thanks for your work @ppkale1995-cimpress! I don't recommend to expire failed jobs on your own, it's better to use the `AutomaticRetry(OnAttemptsExceeded = AttemptsExceededAction.Delete)` instead as shown in the dashboard. However Dashboard UI shouldn't throw an exception anyway, so these changes are useful.

I've added some changes, now waiting for CI to merge the pull request. The fix will be released with version 1.7.9.

odinserj:
The update is available in the version `1.7.8-build-03073` available in the NuGet CI feed – `https://ci.appveyor.com/nuget/hangfire`

ppkale1995-cimpress:
I see the updated version only for `Hangfire.AspNetCore`, `Hangfire.SqlServer.MSMQ`, `Hangfire.NetCore`.
The update is not available for `Hangfire.Core` and `Hangfire.SqlServer`.
Could you make the update available for all the packages as we use only `Hangfire.Core` and `Hangfire.SqlServer`?

ppkale1995-cimpress:
The latest stable version for `Hangfire.Core` is 1.6.13 and for `Hangfire.SqlServer` is 1.6.22 in `https://ci.appveyor.com/nuget/hangfire`

odinserj:
I see both packages of version `1.7.8-build-03073` are available on that feed. Please try upgrading `Hangfire.SqlServer` first by running the following command in your [Package Manager Console](https://docs.microsoft.com/ru-ru/nuget/consume-packages/install-use-packages-powershell#opening-the-console-and-console-controls) window. In this case `Hangfire.Core` should be automatically updated as well.

```powershell
Update-Package Hangfire.SqlServer -Version 1.7.8-build-03073
```

Also you can simply visit the [Artifacts page](https://ci.appveyor.com/project/odinserj/hangfire-525/build/artifacts) of that build and download the required packages. There's a `zip` archive that contains all the assemblies. You can unzip it to the corresponding sub-folders of your `packages` folder.

ppkale1995-cimpress:
I am getting the below error when I use the Update-Package to update `Hangfire.SqlServer`

Update-Package : Unable to find a version of 'Hangfire.Core' that is compatible with 'Hangfire.SqlServer 1.7.8-build-03073 constraint: Hangfire.Core 
(= 1.7.8-build-03073)'.

The dlls in zip worked. Thanks.

PR #1600, opened 2020-01-09
Add support for specifying a long-form HTML job description on the job details page with the JobDescriptionAttribute

benelliott:
Hi,

This PR adds support for "job descriptions", which are long-form blocks of HTML prose which offer more information on a particular job than the name field can allow. If present, the description field is displayed on the Job Details page of the dashboard.

The attribute itself is symmetrical to the JobDisplayNameAttribute. Likewise, I have added DashboardOptions.DescriptionFunc to override the description provider if desired.

Thanks


src/Hangfire.Core/Dashboard/HtmlHelper.cs
```diff
@@ -177,6 +177,43 @@ public string JobName(Job job)
 
             return job.ToString();
         }
+        
+        public string JobDescription(Job job)
+        {
+            if (job == null)
+            {
+                return Strings.Common_CannotFindTargetMethod;
```
pieceofsummer: I believe you can safely return `null` here. There's no need to duplicate "Cannot find target method" in description.

benelliott: Thanks, that's been changed

benelliott:
Hi @pieceofsummer, is there anything else I can do before this PR can be merged?

pieceofsummer:
I’d say it would be reasonable to move `ResourceType` property and resource manager cache to a common base class for all JobDisplayName-like attributes, so there’s no code duplication and wasted memory.

benelliott:
@pieceofsummer Cool - have updated the PR to address this.

src/Hangfire.Core/JobDescriptionAttribute.cs
```diff
@@ -0,0 +1,20 @@
...
+
+namespace Hangfire
+{
+    /// <summary>
+    /// Specifies an HTML description for a job method.
+    /// </summary>
+    [AttributeUsage(AttributeTargets.Method, AllowMultiple = false)]
+    public class JobDescriptionAttribute : JobMetadataAttribute
+    {
+        public JobDescriptionAttribute(string metadataValue) : base(metadataValue)
```
pieceofsummer: Use `description` for the argument name

src/Hangfire.Core/JobMetadataAttribute.cs
```diff
@@ -0,0 +1,71 @@
...
+        /// Gets the metadata for this attribute.
+        /// </summary>
+        public string MetadataValue { get; }
+
+        /// <summary>
+        /// Gets or sets resource type to localize the metadata string.
+        /// </summary>
+        public Type ResourceType { get; set; }
+
+        public string Format(DashboardContext context, Job job)
```
pieceofsummer: Make `Format` virtual (as it used to be)

src/Hangfire.Core/JobMetadataAttribute.cs
```diff
@@ -0,0 +1,71 @@
...
+            if (string.IsNullOrEmpty(metadataValue))
+                throw new ArgumentException("Metadata value is empty", nameof(metadataValue));
+
+            MetadataValue = metadataValue;
+        }
+
+        /// <summary>
+        /// Gets the metadata for this attribute.
+        /// </summary>
+        public string MetadataValue { get; }
```
pieceofsummer: Should probably also make `MetadataValue` protected? Derived classes expose it via their own properties anyways.

benelliott:
Thanks @pieceofsummer, I've made those changes.

odinserj:
Hi @benelliott, great work on this PR! I'm afraid we shouldn't allow the page to display non-sanitized HTML there. Job descriptions themselves are written by developers, but since it's possible to use String.Format-like patterns, it's possible that non-validated user input will be displayed, allowing XSS attacks. 

We could use the [HtmlSanitizer](https://www.nuget.org/packages/HtmlSanitizer/) package to handle this case, but it depends on [AngleSharp](https://www.nuget.org/packages/AngleSharp/) package that's too heavy – 1.23Mb – to be included with Hangfire.Core for just a single miscellaneous purpose.

On the other hand, we can make the Job Details page to be more extensible and allow third-party code to write HTML output after job snippet and before state list. This feature will be extremely useful for different extensions, like batches or even Hangfire.Console so there's no need to use JavaScript to inject some additional HTML.

And if such a feature is ready, we can make job descriptions as extension package that uses HtmlSanitizer on its own, as well as create many other useful extensions. What do you think?

pieceofsummer:
@odinserj I believe descriptions can use markdown formatting instead. Not that you need full HTML for job description, and markdown practically became an industry standard. Also many markdown packages are rather lightweight.

benelliott:
@odinserj Good point - it hadn't occurred to me that user input could be propagated through here(!). I think the extension idea would be interesting but probably a much bigger piece of work than the initial PR - maybe I could have a go if I get some free time soon :) 

In the meantime I like @pieceofsummer's idea of using Markdown instead - it would also make the descriptions more readable in the codebase than HTML anyway.

odinserj:
Fortunately or unfortunately markdown allows HTML to be inlined, and [MarkdownSanitizer](https://www.nuget.org/packages?q=markdownsanitizer) doesn't exists yet, so things will be much more difficult. But with the simple following class we can use HTML, Markdown or even DOC and PDF – but in an extension.

```csharp
namespace Hangfire.Dashboard
{
    public static class JobDetailsRenderer
    {
        public static readonly List<Func<RazorPage, JobDetailsDto, NonEscapedString>> Renderers = 
            new List<Func<RazorPage, JobDetailsDto, NonEscapedString>>();
    }
}
```

```cshtml
@* JobDetailsPage.cshtml *@
@foreach (var renderer in JobDetailsRenderer.Renderers)
{
    @renderer(this, job)
}
```

pieceofsummer:
Actually [MarkdownDeep](https://www.toptensoftware.com/markdowndeep/) claims to have HTML sanitizer :)

odinserj:
Oh thanks, didn't know about MarkdownDeep, looks very interesting. But with that list we can show so many useful things like batches, recurring jobs, job filters, **job descriptions (!)**, tenant information, etc., etc. Custom components are also great, but much more effort is required.

benelliott:
@odinserj Which direction would you like me to take?

odinserj:
@benelliott, just have a rest now. I will implement that list with additional renderers and let you know once it's ready. There's no problem to have it in 1.7.10, because it's dashboard feature, and it's simple enough to wait for 1.8.0. Once it's ready, everything is simple.

As I've said before, we can't ignore possible XSS vulnerability and implement this PR as is. Additional package like `MarkdownDeep` brings additional license, and shouldn't be added in patch versions. And anyway, I don't like to add the whole new package for a single reason.

benelliott:
Ok thank you!

odinserj:
With Hangfire 1.8.0 it will be possible to add custom widgets to the Job Details page, including some helpers that read data from an attribute applied to a method and displays additional information on the page, please see https://www.hangfire.io/blog/2023/03/30/hangfire-1.8.0-rc3.html#custom-renderers-on-the-job-details-page for details.

PR #1630, opened 2020-03-06
Add PrefixPath in DashboardOptions.

shenjielx:
image
add prefix path in url.


odinserj:
Thanks @shenjielx, I'm merging the PR and it will be available in Hangfire 1.7.10 shortly.

PR #1638, opened 2020-04-03
Add 'Clock' for UtcNow injection

NooNameR:
It could be really helpful while testing: you can inject your own `IClock` and control how jobs are executing.


hankovich:
Hi @odinserj, are there any plans to merge this PR? Or to provide another way to replace the system clock with a custom one?

WhatzGames:
Hi, just wanting to give my two cents on these somewhat older changes. It might be worth considering to use the official abstract TimeProvider now. It also has a BCL package.

PR #1648, opened 2020-04-18
default datetime comes from TimeZoneInfo

fahrigoktuna:
Hi Odin,

Your RecurringJobManager constructors works with only DateTime.UtcNow, not affected by has been sent TimeZoneInfoResolver,

For Example:
*******
new RecurringJobManager().AddOrUpdate(String.Format("{0}:{1}", schedulejob.Code, schedulejob.Id), job, schedulejob.RunTimePattern,TimeZoneInfo.Local, queuename);
*******
If developers create recurring job (shown at above), TimeZoneId updated/created succesfully however, CreatedAt and NextExecution has values updated/created with DateTime.UtcNow,

This pull request solves that issue, your constructors now able to resolve datetimeformat via ITimeZoneResolver.

### DIFF
--- a/src/Hangfire.Core/ITimeZoneResolver.cs
+++ b/src/Hangfire.Core/ITimeZoneResolver.cs
@@ -21,15 +21,21 @@ namespace Hangfire
 {
     public sealed class DefaultTimeZoneResolver : ITimeZoneResolver
     {
+        private TimeZoneInfo _timeZoneInfo;
         public TimeZoneInfo GetTimeZoneById(string timeZoneId)
         {
-            return TimeZoneInfo.FindSystemTimeZoneById(timeZoneId);
+            _timeZoneInfo = TimeZoneInfo.FindSystemTimeZoneById(timeZoneId);
+            return _timeZoneInfo;
         }
+
+        public TimeZoneInfo GetCurrentTimeZoneInfo() => _timeZoneInfo;
     }
 
     public interface ITimeZoneResolver
     {
         [NotNull]
         TimeZoneInfo GetTimeZoneById([NotNull] string timeZoneId);
+
+        TimeZoneInfo GetCurrentTimeZoneInfo();
     }
 }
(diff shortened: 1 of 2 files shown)


fahrigoktuna:
AppVeyor pipeline failed because of  System.InvalidOperationException : There is already an open DataReader associated with this Command which must be closed first.

odinserj:
Yes, all the date time values are stored in UTC, and time zone resolver acts only when calculating next occurrences by the corresponding Cron expression. 

So everything is expected and I don’t understand what problem you are trying to solve.

fahrigoktuna:
If developer create recoring job with passing timezone for instance timezonelocal which is turkey standart time has values will be store timezoninfo turkey standart however created at and next execution date will be stored as utc(next execution is cron based utc time) so, backgground server will  take schedule job from queue when exact timezone however dates are stored utc time not timezonelocal time.

Lets explain with an example,

Recurringjob send to hashes with cron format every midnight 00:00 recurringjobmanager saves to hash table with nextexecution time 21:00 it will appear hash table however willl be executed 00:00 since timezone info +3 turkey standart time, above code helps to store nextecution time 00:00 in hash table.

Hope, I explain clearly

odinserj:
Everything is stored in UTC, right. But when Hangfire calculates next occurrence, it takes UTC date, and passes to the Cronos library together with the time zone. Cronos library then convert the given UTC time to the local one using the given time zone, calculates next occurrence based on the given Cron expression and local time, and then converts it back to UTC time to store it in a database and protect values from being specific from server’s current time zone.

Are you using a custom type resolver?

fahrigoktuna:
I'm closing this PR. I'm just trying to say that; Created schedule Job : 
TimeZoneId	Turkey Standard Time

Cron	58 22 * * *

the job will be processed at 22:58 at local Turkey time. This is how your code works and nothing wrong with that. However, CreatedAt and NextExcution Time stores as UTC time which -3 hours from local time. 

What it can confuse developers mind. If I set time zone local or whatever I setted. CreatedAt and NextExecution date will be stored as my setted TimeZone Ithink.

Nevertheless, this is not affect queue processing. Cron set 22:28 with local time zone. So When 22:28 at my server it will be process. I would like to see that Create at and NExtExecution time setted as my local time zone.

Thanks for you comments.

odinserj:
It’s stored in UTC to prevent wrong behavior when local time zone is changed, or when daylight saving time is applied/unapplied, or when rules are changed within a time zone itself. Local time is a volatile value, and we shouldn’t depend on it.

PR #1651, opened 2020-04-23
all dashboards requests will include initial query string

Alexei000:
If initial query to dashboard (i.e. `/hangfire`) also provides a query string, this will be used for all links and requests within the dashboard. 

This is useful to make the dashboard accessible based on a bearer token or similar. See [this discussion](https://discuss.hangfire.io/t/using-bearer-auth-token/2166).

If the query string is empty (default case), no link / url is changed.


odinserj:
Hi @Alexei000, this is an interesting idea to make everything work with SPA. But I have a question related to security – do I understand everything correctly, and some other person who can see the URL and that query string may use it to access the Dashboard UI?

Alexei000:
Hello, @odinserj! Indeed this is an issue, although I expect the string not to be entirely visible. This should be regarded more as a workaround until a better solution is found. 

Thanks.

PR #1656, opened 2020-04-29
[Dashboard] Adding Fullwidth support

PaulARoy:
This derives from the awesome work of @JesseKlaass on this PR: https://github.com/HangfireIO/Hangfire/pull/937

It focuses only on full width support with a `UseFullWidth` option on `DashboardOptions`.

Things to consider:
- `UseFullWidth` is enabled by default; this could be the desired option or not. It can easily be changed
- The support for job pages uses a specific css class to avoid a huge 25% sidebar (by blocking with max-width). It's not perfect but it keeps the same classic and mobile rendering, and renders correctly for full-width (even if it's not perfect). That explains the number of files modified

I am open to suggestions and improvements if you want!


odinserj:
Thanks a lot @PaulARoy, really great change! I've scheduled this PR to Hangfire 1.7.12, since changes are small enough.

odinserj:
We also need to add @JesseKlaasse to the authors of this feature, perhaps with the `Co-Authored-By` or like commit message when merging the PR.

yuzd:
i really like UseFullWidth

penenkel:
This feature is interesting, but seems to be stalled. Any reason for this? @odinserj

ViolentRage:
> This feature is interesting, but seems to be stalled. Any reason for this? @odinserj

Actually I can't find it in DashboardOptions class, using version 1.7.24

odinserj:
Full width layout will be enabled by default in Hangfire 1.8.0, please see https://www.hangfire.io/blog/2023/03/30/hangfire-1.8.0-rc3.html#full-width-support for details. There were problems with sidebar size/positioning that required extra work here, but now everything is working fine there. Thank you for the great idea @PaulARoy!

PR #1659, opened 2020-05-01
[Dashboard] Display job arguments

PaulARoy:
This derives from the awesome work of @JesseKlauss on this PR: #937

It focuses only on displaying arguments of a job in all lists with a `DisplayArgumentsInLists` option on `DashboardOptions`. Set to false by default.

Translations included.


odinserj:
Hm, this is an interesting feature, but I don't understand how long arguments are displayed here without breaking the whole table? Also, it's possible to include background job arguments by using the `DisplayName` attribute:

```csharp
[DisplayName("MyMethod: {0} arg1, {1} arg2")]
public string MyMethod(int arg1, string arg2, string json) {}
```

So there's a big question regarding this feature.

PaulARoy:
I honestly didn't know about the display name, but I still think it's an interesting feature to activate for all jobs. As it's not necessarily needed for everyone it's better to deactivate by default, but it's a feature I'd really find relevant for most of my jobs.

The long arguments is indeed an issue. As documentation say, we should keep job arguments as simple as possible, but it still can happen. The fullwidth feature is essential for this to work in my opinion.

But let me work on the long arguments problem; I'll try to find a satisfying solution.

PR #1663, opened 2020-05-06
[Dashboard] AspNetCore 3.0+ authorization policy support via extensions

dasiths:
The current DashboardOptions constructor sets the `LocalRequestsOnlyAuthorizationFilter` as a default filter. When you setup your Hangfire Dashboard using the new endpoints pattern and set an authorization policy it still goes ahead and uses the default anyway. To circumvent it we need to pass an empty list to `Authorization` as shown below.

```
            app.UseEndpoints(endpoints =>
            {
                endpoints.MapControllers()
                    .RequireAuthorization();
                endpoints.MapHangfireDashboard("/dashboard/hangfire", new DashboardOptions()
                    {
                        // We need to pass an empty list or Hangfire will use a local authorization filter
                        Authorization = new List<IDashboardAuthorizationFilter>()
                    })
                    .RequireAuthorization("hangfire-specific-auth-scheme");
            });
```

The new extension method introduced in this PR supports passing an authorization policy name and removes the `LocalRequestsOnlyAuthorizationFilter` from the default if no `DashboardOptions` instance has been passed.

### DIFF
--- a/src/Hangfire.AspNetCore/HangfireEndpointRouteBuilderExtensions.cs
+++ b/src/Hangfire.AspNetCore/HangfireEndpointRouteBuilderExtensions.cs
@@ -1,5 +1,5 @@
 #if NETCOREAPP3_0
-
+using System.Collections.Generic;
 using Hangfire.Annotations;
 using Hangfire.Dashboard;
 using Microsoft.AspNetCore.Builder;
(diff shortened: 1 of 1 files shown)


dasiths:
How can I get this reviewed?

src/Hangfire.AspNetCore/HangfireEndpointRouteBuilderExtensions.cs
```diff
@@ -47,6 +47,35 @@ public static class HangfireEndpointRouteBuilderExtensions
...
+            [NotNull] string authorizationPolicyName,
+            [NotNull] string pattern,
+            [CanBeNull] DashboardOptions options = null,
+            [CanBeNull] JobStorage storage = null)
+        {
+            if (options == null)
+            {
+                options = new DashboardOptions()
+                {
+                    Authorization = new List<IDashboardAuthorizationFilter>() // We don't require the default LocalHost only filter since we provide our own policy
```
leonio, line 71: Perhaps this can just be `Authorization = Enumerable.Empty<IDashboardAuthorizationFilter>()`

dasiths:
@odinserj any feedback on this?

src/Hangfire.AspNetCore/HangfireEndpointRouteBuilderExtensions.cs
dasiths: Thanks for the review @leonio. It doesn't make a whole lot of difference for readability or performance IMO. Can we let this go as is?

leonio: I don't have rights sorry, however this MR was very helpful for an issue I was resolving. FWIW hope it makes it through a full review.

odinserj:
Thanks for your work and patience! Will merge this pull request and release changes with 1.7.24.

PR #1715, opened 2020-07-27
#7895: Hangfire Dashboard does not know job names of missing assemblies

rolfik:
I have started to use Hangfire for some of our applications. Thank You for the work.
One of applications is UWP one which cannot host Hangfire Dashboard itself. So I have created Windows service Hangfire Dashboard server monitoring the application externally via shared SQLite database.
To see actions in the Dashboard instead of "Missing type etc." messages, I have augmented it to load akction information from InvocationData when Type cannot be found/loaded.
This allows me to manage jobs even without action assemblies.

### DIFF
--- a/appveyor.yml
+++ b/appveyor.yml
@@ -7,7 +7,7 @@
 #   - Section names should be unique on each level.
 
 # Don't edit manually! Use `build.bat version` command instead!
-version: 1.7.11-build-0{build}
+version: 1.7.12-build-0{build}
 
 os: Visual Studio 2019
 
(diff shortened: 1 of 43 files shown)


rolfik:
I see the conflict with my solution with newly added DBG properties which try to solve similar thing, but only on Job Details page.
More discussion is required how to solve the conflict.

odinserj:
Unfortunately I see breaking changes in JobName method and it will not work without InvocationData argument – NullReferenceException will be thrown, so we can’t simply add method overload and pass null. And I also see InvocationData class is changed itself and that’s dangerous – did you try to run unit tests at least for Hangfire.Core? AppVeyor needs merge conflicts to be resolved first to see whether all the unit tests passed.

It’s not hard to build a feature, it’s hard to build it in a way which will not break processing in anyone’s environment. Breaking changes are the simplest problems to find, but there are much harder ones – forward (i.e. older versions can process data generated from the newer ones, for making rolling upgrades possible) and backward (vice versa) compatibility.

PR #1737, opened 2020-09-07
support for configuring UserSerializerSettings via callback

Yaevh:
As of 1.7.12, there's no way to modify existing `UserSerializerSettings`, they can only be replaced entirely. This PR adds `IGlobalConfiguration.UseSerializerSettings(Action<JsonSerializerSettings> settingsConfiguration)` extension method to enable users to modify existing settings without having to write them from scratch. Also contains `IGlobalConfiguration.UseRecommendedSerializerSettings(Action<JsonSerializerSettings> settingsConfiguration)` to allow using recommended settings, but with user-defined tweaks.

### DIFF
--- a/src/Hangfire.Core/Common/SerializationHelper.cs
+++ b/src/Hangfire.Core/Common/SerializationHelper.cs
@@ -307,7 +307,7 @@ namespace Hangfire.Common
             }
         }
 
-        private static JsonSerializerSettings GetUserSerializerSettings()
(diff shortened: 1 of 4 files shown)


odinserj:
Thanks @Yaevh, I've added some changes to avoid breaking ones (see the `CleanSerializerSettingsAttribute` file, it was required to specify type before the new changes), and limited the scope of a new parameter to apply it only in the `UseRecommendedSerializerSettings` method – I don't want to touch implicit user settings in any way.

PR #1772, opened 2020-11-30
German (Deutsch) Dashboard localization

do-it-ger:


odinserj:
Thank you @do-it-ger! I've removed some spaces from HTML tags and now merging the change.

PR #1775, opened 2020-12-08
Use constant value for StateName

u1-liquid:
Use constant value for StateName.

With this, `StateClass.StateName` could be used in switch-case statement.


Frazi1:
`static readonly` fields are evaluated at runtime when `const` — at compile time.

If state names are constants and if at some point a constant value is changed, it will require every Hangfire consumer to rebuild their code for these changes to take effect.

On the other hand, consumers will not have to do anything if it is a readonly field.

u1-liquid:
The state names are already stored into storage constantly, so if some point that value is changed, it means other consumers couldn't fetch jobs correctly until rebuild their codes for those changes to take effect.

I think there is no advantages about this point to keep `StateName` field to `static readonly`.

Also, this changes only effects to `Hangfire.Core`, So basically users who uses this library will NOT changes this part of codes,
 If they need to change this kind of values, the state class should be defined in their code, which is not affected by this change.

> 
> 
> `static readonly` fields are evaluated at runtime when `const` — at compile time.
> 
> If state names are constants and if at some point a constant value is changed, it will require every Hangfire consumer to rebuild their code for these changes to take effect.
> 
> On the other hand, consumers will not have to do anything if it is a readonly field.

PR #1800, opened 2021-01-27
Add an overload for AddHangfireServer utilizing IServiceProvider

penenkel:
This allows the configuration of `BackgroundJobServerOptions` to be based on other services, options, etc..

While there exists the workaround of registering the `BackgroundJobServerOptions` manually
```
services.AddSingleton(provider => {
var other = provider.getRequiredService<OtherOptions>();
return new BackgroundJobServerOptions
{
    Queues = new[] { "default", "misc", other.SomeQueueName},
    WorkerCount = 1,
};
});
services.AddHangfireServer();
```
This fails if one wants to register multiple HangfireServer with differing configurations (all based on some other service/options).

### DIFF
--- a/src/Hangfire.AspNetCore/HangfireServiceCollectionExtensions.cs
+++ b/src/Hangfire.AspNetCore/HangfireServiceCollectionExtensions.cs
@@ -139,6 +139,24 @@ namespace Hangfire
 
             return services;
         }
+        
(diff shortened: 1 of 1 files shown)


odinserj:
Thanks @penenkel, will release this in version 1.7.20 today.

PR #1803, opened 2021-02-01
Support async DashboardAuthorization in AspNetCore for Hangfire 1.x

rosenbjerg:
Safe and backwards compatible support for async dashboard authorization in when using AspNetCore dashboard middleware.
Only negative thing is that the `AsyncAuthorization` prop can even when not using AspNetCore

I'm not sure if this is intentionally not supported?


rosenbjerg:
Both failing checks seems to fail due to test infrastrucute problems

odinserj:
Awesome @rosenbjerg, thanks for your patience. Will be available in 1.7.24.

PR #1853, opened 2021-05-04
Replace deprecated PerformContext ctor usage

kumaheiyama:
The `PerformContext(IStorageConnection, BackgroundJob, IJobCancellationToken)` constructor is deprecated, but used in the `PerformContextMock` class. Code does build and run, but throws an error when running a code validator like SonarQube.

Update code to use other ctor and add mock for `JobStorage`


odinserj (approved):

odinserj:
Thanks @kumaheiyama!

PR #1861, opened 2021-05-16
Added JobId property to JobPerformanceException for error logging purposes

Plasma:
## Overview
This PR adds a `string JobId` property to the `JobPerformanceException` class so that error logging code (exception handlers, etc) can access the `JobId` that was running that caused this issue.

## Notes
Initially I was going to provide a `BackgroundJob Job` property instead of a `string JobId` property, but that would not let the exception class be `Serializable` any more.

There is also a few `Obsolete` call code paths I do not pass in the `JobId` so this property could be `null`.


odinserj:
Thanks @Plasma! Will release these changes as a part of 1.7.23 release.

PR #1869, opened 2021-06-03
  Implemented fixed-top behaviour for error alerts

danillewin:
Before: alerts are not visible after scrolling, sidebar has excess margin

image

image

After: alerts are fixed to top and always visible. Sidebar has proper margin

fixed-1

fixed-2


odinserj:
Oh-oh, that's really nice! Looks like I'm now able to display other error messages :tada:

PR #1878, opened 2021-06-14
Avoid memory leak in AppBuilderExtensions

### LINKED ISSUE #1877: AppBuilderExtensions.Servers leaks BackgroundJobServer instances
AppBuilderExtensions.Servers is never cleaned up, so it leaks BackgroundJobServer instances.

LordJZ:
Since ConcurrentBag does not support removal of a specific object, and there is no thread-safe set class, here ConcurrentDictionary is used as a Set. Fixes #1877


odinserj:
Thanks @LordJZ!

PR #1887, opened 2021-06-24
Implemented back-to-site link responsiveness

danillewin:
Make back-to-site button from image below collapse, for saving space on narrow screens like at the last image of the PR

**Default**

back-to-site 1

**Collapsed at 768-992px**

back-to-site 2


odinserj:
Thanks @danillewin!

PR #1888, opened 2021-06-24
Remove excess paddings of alert

danillewin:
Implemented fixed-top behaviour for error alerts 

**Before**

**After**


odinserj:
Awesome @danillewin, thanks!

PR #1893, opened 2021-07-05
Implemented full-width support

danillewin:
Before:
 
 Fixed width, even when screen width is wider than 1200px
 
 
full-screen-before

After:

For screens of all sizes container width is now relative.

**Wider than 1200px**

full-width-L

**992px - 1200px**

full-width-M

**768 - 992px**

full-width-S

**Narrower than 768px**

full-width-XS


odinserj:
Awesome @danillewin, brilliant work!

PR #1895, opened 2021-07-06
Moved from inline style to classes at Job's state indication

danillewin:
There's someplaces, where we indicate Job's state through colors: labels and state cards

1. Job Details Page

image

2. Recurring Jobs Page

image

3. Awaiting Jobs Page 

4. Retries Page 

image

Problem: colors are applied by `styles` attribute, which is not a good practice and blocks potential features, e.g. dark theme.

Solution: move from inline style to modification classes, e.g. `.state-card_state_processing`, `.label_state_awaiting`


odinserj:
Closing this as obsolete, this feature was implemented in the https://github.com/HangfireIO/Hangfire/commit/0c816995ab480735a7d560abd131b6875f6c312b commit.

PR #1916, opened 2021-08-03
Fix incorrect validation for HeartbeatInterval

GitHubPang:
Fixed incorrect argument validation in `HeartbeatInterval` setter, which was checking against the wrong constant.

Actual behaviour is unchanged since their values are identical.


odinserj:
Good catch @GitHubPang, thanks!

PR #1965, opened 2021-11-09
Fix 2 typos in German translation

saxx:
Only very minor fix: Plural of "Auftrag" in German is "Aufträge" (not "Auftrage").


odinserj (approved):
Great, thanks a lot for handling this! :heart:

PR #1979, opened 2021-12-07
Fixing translations in Turkish translation

can-zengin:
Some of the translations were hard to understand.


odinserj:
Thank you for the update! Maybe we should also ask the original author of Turkish translation @csarigul if he has some time to review the changes.

odinserj (approved):

PR #2007, opened 2022-03-04
add try-catch DistributedLockTimeoutException means another Hangfire …

shenjielx:
…server did this work.

### DIFF
--- a/src/Hangfire.Core/Server/RecurringJobScheduler.cs
+++ b/src/Hangfire.Core/Server/RecurringJobScheduler.cs
@@ -232,29 +232,41 @@ namespace Hangfire.Server
             string recurringJobId,
             DateTime now)
         {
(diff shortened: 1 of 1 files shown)


odinserj:
@shenjielx could you please add more details about the problem you observe without this change? There's an outer distributed lock acquisition to synchronise recurring job scheduler executions on different machines so it's unlikely due to another server.

shenjielx:
> @shenjielx could you please add more details about the problem you observe without this change? There's an outer distributed lock acquisition to synchronise recurring job scheduler executions on different machines so it's unlikely due to another server.

But I want to run in different machines, the UseConnectionDistributedLock in EnqueueNextRecurringJobs method, i set the lock only in same machine. When I run more recurring jobs (1 second), I found it spent time more than 1 second, it's sync every recurring job, so I want to run diff recurring job in different machines. I can?

PR #2060, opened 2022-07-06
OnFailActionsSupportAttribute.cs

dyha89:
odinserj, please tell what do you think about such feature to run jobs on other job failure

i have seen few people asking about it, and I found it useful in my case too
maybe it's a feature for separate nuget or something but I wanted to ask you first about contribution

i read that your concept is that if job has failed, that means that developer needs to deploy fix and then restart the job
but in my scenario i have found usecases, where job has failed but we can do some automatic system reparation stuff that may bring it to life and then restart the job automatically

for example:
we are sending some email
var sendJob = client.Enqueue(() => SendEmail());
but if all retries has failed, maybe something really bad happened with the network or somehting, and we can for example restart network interfaces to fix it before some developer can look at the issue
so we are defining such on fail action
var fallback = client.OnJobFail(sendJob, () => RestartNetworkInterfacesAndDoSomeCrazyMagic());
after fallback action success, change state of our failed job to enqueued again to give it a try
client.ContinueJobWith(fallback, () => client.ChangeState(sendJob, new EnqueuedState(), FailedState.StateName));

it's not tested well yet, but I wanted to post it first as a draft pull reqest to ask if you see it useful

i have based this on your ContinuationsSupportAttribute class

### DIFF
--- a/src/Hangfire.Core/OnFailActionsSupportAttribute.cs
+++ b/src/Hangfire.Core/OnFailActionsSupportAttribute.cs
@@ -1,0 +1,321 @@
+using System;
+using System.Collections.Generic;
+using System.Linq.Expressions;
+using System.Threading.Tasks;
+using Hangfire.Annotations;
+using Hangfire.Common;
+using Hangfire.States;
+using Hangfire.Storage;
+
+namespace Hangfire
+{
+    // POC: odinserj, please tell what do you think about such feature to run jobs on other job failure
+
+    // i have seen few people asking about it, and I found it useful in my case too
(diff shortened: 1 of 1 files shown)


odinserj:
Thanks for your work. Version 1.8.0 will add the Exception field to the Deleted state and will allow to specify continuations for the deleted state also (which will act as a final state for failed jobs). There will be `JobContinuationOptions.OnlyOnDeletedState` option and also ability to specify multiple values for the continuation options. I will add more release notes a bit later, now changes are available in the form of release notes, please see https://github.com/HangfireIO/Hangfire/releases/tag/v1.8.0-rc1 for details.

PR #2064, opened 2022-07-22
Changed service registration from Transient to Singleton to able consume it

krzysiek-b:
Seems like it makes more sense to register the service as Singleton to allow any consumer inject it into its classes rather than having this transient and not being able to consume it at all.


odinserj:
Unfortunately this will break existing code, where multiple background job servers are started in the same application that's an existing use case. Different applications, different environments, as far as I remember `BackgroundJobServerHostedService` class is public so you can use your own registrations directly instead of calling the `AddHangfireServer` method.

PR #2092, opened 2022-09-21
Swedish translation

karl-sjogren:
Adds a first version of Swedish translation of the dashboard.


odinserj:
Thank you so much! Thrilled to see even more translations!

odinserj (approved):

PR #2111, opened 2022-10-24
Document the exception for the Job Cancellation Token

judah4:
Updates the `JobCancellationToken` to include intellisense hints for the exception it throws for `ThrowIfCancellationRequested()`.

These comments were taken from CancellationToken comments.

### DIFF
--- a/src/Hangfire.Core/JobCancellationToken.cs
+++ b/src/Hangfire.Core/JobCancellationToken.cs
@@ -31,18 +31,7 @@ namespace Hangfire
 
         public static IJobCancellationToken Null => null;
 
-        /// <summary>
-        /// Throws a <see cref="System.OperationCanceledException">OperationCanceledException</see> if
-        /// this token has had cancellation requested.
-        /// </summary>
-        /// <remarks>
-        /// This method provides functionality equivalent to:
-        /// <code>
-        /// if (token.ShutdownToken.IsCancellationRequested)
-        ///    throw new OperationCanceledException(token);
-        /// </code>
-        /// </remarks>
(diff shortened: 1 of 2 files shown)


src/Hangfire.Core/JobCancellationToken.cs
```diff
@@ -31,6 +31,18 @@ public JobCancellationToken(bool canceled)
...
+        /// this token has had cancellation requested.
+        /// </summary>
+        /// <remarks>
+        /// This method provides functionality equivalent to:
+        /// <code>
+        /// if (token.ShutdownToken.IsCancellationRequested)
+        ///    throw new OperationCanceledException(token);
+        /// </code>
+        /// </remarks>
+        /// <exception cref="System.OperationCanceledException">The token has had cancellation requested.</exception>
```
odinserj, line 45: I think we can use the [`<inheritdoc />`](https://learn.microsoft.com/en-us/dotnet/csharp/language-reference/xmldoc/recommended-tags#inheritdoc) tag to avoid duplication here. As far as I understand in this case documentation will be taken from the `IJobCancellationToken` interface.

odinserj (changes requested):
Thanks a lot for polishing this! I think we can use the `<inheritdoc />` tag to avoid duplication as I've commented above and then merge this PR.

src/Hangfire.Core/JobCancellationToken.cs
judah4: Oh, I didn't know about this, that's cool! I went ahead and made the change.

odinserj (approved):
Thanks @judah4, everything is perfect now!

PR #2129, opened 2022-11-12
first two thirds of transated text is prepared

YanStan:


YanStan:
need git pull & finished work first

Issue #1779, opened 2020-12-10
Why are the Interval functions being removed from Cron.cs?


RichardNobel:
Why are the `MinuteInterval`, `HourInterval`, `DayInterval `and `MonthInterval` functions considered Obsolete?
> "Please use Cron expressions instead. Will be removed in 2.0.0"

In my opinion, it makes the code easier to read with `MinuteInterval(15)` instead of a string notation `"*/15 * * * * "`.

Of course I can copy these (or create similar) functions in my own project/library.

But is there a specific reason why they need to be removed from Hangfire.Core?

BenJenkinson:
Hi @RichardNobel,

Issue #1041 describes the reason they were marked as obsolete. I've quoted it below:

> Neither of these methods works correctly, when the specified interval is not a multiple of the corresponding field's range length (max-min). For example, when we are using `Cron.HourInterval(10)`, we will not get the documented every 10 hours semantic, instead we'll get the following occurrences:
> 
> ```
> Nov 6, 00:00
> Nov 6, 10:00
> Nov 6, 20:00 
> Nov 7, 00:00
> ```
> 
> Time interval between the last two entries is only 4 hours and violates what's written in the method's name and in its comments. Cron expressions don't allow to specify "every N" recurring intervals in general, and other methods should be used instead of confusing methods.

Issue #1054, opened 2017-11-10
Add proper support for time intervals in recurring jobs


odinserj:
Since `Cron.XXXInterval` methods should be deprecated as per #1041, a question arises – how to handle intervals then? One could use `ScheduledState` to create a job and re-schedule after each completion. But this is not an obvious option, and could lead to much more non-obvious consequences like 5 retries or an original job caused 5 scheduled jobs to be created instead of 1 with a simple implementation.

Instead, recurring jobs should be modified to support the following recurring schemes (yes, there will be no support for seconds yet). This is just to handle simple cases, with a simple implementation. Since recurring job is a has, we could save backward compatibility without any problems.

1. Every X minutes
2. Every X hours at MM minutes
3. Every X days at HH:MM
4. Every X months at DD day at HH:MM

crawdell:
I'd like to propose something I've done before to handle cron intervals. I needed a solution to "every 2 weeks" when using Ncrontab Advanced and so I extended the functionality to allow for "Xnn" to be added to a cron expression.

Example:
00 00 * * FRI X2

Without "X2" this would just be every Friday at midnight. But with X2 I would getnextoccurance 2 times and use the second value as the next run time which would be 2 weeks.  
This can be useful for doing all kinds of new variations. It can also lead to confusing results sometimes if you don't think about your expression carefully but that's true of all cron expressions.

This I think helps fill in a gap that cron has and has been very useful for my scheduled jobs. Now that I'm using Hangfire I'll love to see this included.

I'm on the 1.7.0 beta and it works great. If this could somehow be added that would be very helpful. If not I might try my own version with the source either by extending Hangfire or more likely Cronos.

[I posted this in another response as well.  Sorry if this confuses, I didn't see this issue then]

antgel:
Does anybody plan to work on this extremely useful feature? Something like "every x weeks from date y" would be great.

SomeOneIsHereAtGit:
Hi, is there any update on this?

PR #2141, opened 2022-11-24
Update Cron.cs - MinuteInterval & HourInterval

staviloglu:
Updated MinuteInterval & HourInterval methods to make them available (not obselete) again.

Added defensive code against interval value to get rid of the problem described in HangfireIO/Hangfire#1041 

@odinserj 

Also related:
HangfireIO/Hangfire#1054
HangfireIO/Hangfire#1779

### DIFF
--- a/src/Hangfire.Core/Cron.cs
+++ b/src/Hangfire.Core/Cron.cs
@@ -1,4 +1,4 @@
-﻿// This file is part of Hangfire. Copyright © 2013-2014 Hangfire OÜ.
+// This file is part of Hangfire. Copyright © 2013-2014 Hangfire OÜ.
 // 
 // Hangfire is free software: you can redistribute it and/or modify
 // it under the terms of the GNU Lesser General Public License as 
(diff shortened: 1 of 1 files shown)


odinserj:
Thanks @staviloglu, this is an interesting way of providing the feature while trying to avoid its potential problems. However I still believe that one day there's much better implementation of this with proper intervals based on duration and not on the Cron format, and I believe sometimes it's much better to leave pain points to have more motivation of fixing them in full.

staviloglu:
@odinserj I totally understand you but also know that many people still use these methods even if they are obselete and they may or may not be informed about the issue #1041. 

This PR, if they check change-logs or update their version of Hangfire, would inform anyone who uses these methods. 

I need to confess that I still want to use those methods with the proper values  😄  

Anyway, thanks for your time and this awesome library, keep up the good work.

Spiralis:
Well, there's always this library: https://github.com/schyntax/cs-schyntax

Used this many years ago and loved the system. It is ported to various languages, all based on Schyntax: https://github.com/schyntax

PR #2161, opened 2023-02-03
Additional query to get jobs by ids

aistisolendra:
The only current way to get jobs by ids is to get chunks of jobs with ScheduledJobs(from, to) function, which is time-consuming if you have many of them. 
This PR tries to solve this issue by providing directly a function that lets you get specific jobs that you need.


burningice2866:
Auch, thats going to break all existing 3rd-party storage providers.

odinserj:
You are correct Pauli, unfortunately can't merge it to 1.X version due to breaking changes.

PR #2181, opened 2023-03-24
Add JobCount(string) method to get count of jobs in any state

Issung:
Currently monitoring API only supports getting job count for the given states (e.g. processing, scheduled, failed)
Useful for developers who create jobs with custom states.


Issung:
Need to fix my gpg

PR #2182, opened 2023-03-24
Add JobCount method to MonitoringApi to get count of jobs in any state

Issung:
Currently monitoring API only supports getting job count for the given states (e.g. processing, scheduled, failed)
Useful for developers who create jobs with custom states.
Signed commits now.

### DIFF
--- a/src/Hangfire.Core/Storage/IMonitoringApi.cs
+++ b/src/Hangfire.Core/Storage/IMonitoringApi.cs
@@ -35,6 +35,7 @@ namespace Hangfire.Storage
         JobList<FailedJobDto> FailedJobs(int from, int count);
         JobList<DeletedJobDto> DeletedJobs(int from, int count);
 
+        long JobCount(string stateName);
         long ScheduledCount();
         long EnqueuedCount(string queue);
         long FetchedCount(string queue);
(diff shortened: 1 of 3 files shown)


Issung:
I realise you have a redis storage and preview memory storage too that this would need implementing on. Given it is a reasonably simple function I hope you can do it quickly and easily, if you would prefer I do it where can I access those classes to make another PR matching this one?

burningice2866:
... and I have a Storage provider for a specific CMS and there are MySQL and mongoddb and many other storage providers out there.

So no changing of public interfaces, it break things.

Qubitus:
I agree with the need for this after having implemented our own custom states recently.
As mentioned, adding this method to the interface would break existing storage provider implementations.
You could avoid this with a default interface implementation, but that's only available in .NETCore 3.0/C# 8.0.

A workaround for now could be to add a default implementation (which just throws a NotImplementedException instance) to the JobStorageMonitor base abstract class.
That way storage providers based on this class can override the method and provide a concrete implementation if they want. Casting the IMonitoringApi to JobStorageMonitor would then allow you to use this method.
Not a perfect solution by any means, but would slightly improve the experience of working with custom states.

Issung:
I like that Idea @Qubitus, I will update the PR soon with that change. 
As for @burningice2866 's comment, breaking changes are a necessity to improve APIs. If Hangfire itself has a breaking change promise, then this change can be held for a major version number release.

PR #2203, opened 2023-05-09
state-card-state-active not very dark

coolhome:
Fix when state-card is active. It retained the light background color. The other card states appear to override to transparent.

Before:
image

After:
image


odinserj (approved):

odinserj:
Great, thank you so much for fixing this!

PR #2320, opened 2023-11-18
Add validation for the server queues

MuhamedAbdalla:
Allow only lowercase, letters, digits and underscores using regex


MuhamedAbdalla:
@odinserj 
I would appreciate if you can take a look

odinserj:
Good idea, sorry for the delay. I think it's better to use the internal `EnqueuedState.ValidateQueueName` method to avoid duplicating the validation logic. I think this will also fix the build error caused by an absent namespace usage statement.

MuhamedAbdalla:
> Good idea, sorry for the delay. I think it's better to use the internal `EnqueuedState.ValidateQueueName` method to avoid duplicating the validation logic. I think this will also fix the build error caused by an absent namespace usage statement.

@odinserj 
Done

odinserj:
Thanks @MuhamedAbdalla! I also considered that some environments may include wrong queue names and correct ones at the same time and work fine at least for the correct ones. If we change logic and will throw an exception, their background job servers will stop working, causing troubles for them.

What if we just emit a warning message during background job server start where explicitly state that some queue names are wrong for the first step? I've committed the corresponding changes to this PR, please review them.

During the next step we can filter out wrong queue names and emit an error message, and pass to workers only correct queue names (so dashboard will also include only correct names).

And only at the next step, after some years, we will be able to throw an exception when wrong queue name is passed.

odinserj (approved):

PR #2324, opened 2023-11-22
Avoid acquiring format for macro expression parsing

MuhamedAbdalla:
Avoid acquiring format for macro expression parsing. As It's not needed to choose a format while parsing macro expression.


odinserj:
Thanks @MuhamedAbdalla, the build is failing, because there's no `StartsWith` overload with `char` type on the `netstandard1.3` platform, there are only `string`-based alternatives. Please try to use the `"@"` instead of `'@'` argument and specify the `StringComparison.OrdinalIgnoreCase` argument for comparison settings there.

Also, I think that dedicated test should be added to ensure that `@hourly`-like parameters are working fine in the `RecurringJobEntity` after doing the changes. If you have any troubles with test please let me know and I will assist you.

MuhamedAbdalla:
> Thanks @MuhamedAbdalla, the build is failing, because there's no `StartsWith` overload with `char` type on the `netstandard1.3` platform, there are only `string`-based alternatives. Please try to use the `"@"` instead of `'@'` argument and specify the `StringComparison.OrdinalIgnoreCase` argument for comparison settings there.
> 
> Also, I think that dedicated test should be added to ensure that `@hourly`-like parameters are working fine in the `RecurringJobEntity` after doing the changes. If you have any troubles with test please let me know and I will assist you.

@odinserj 
Done, Can you recheck it, please?

odinserj (approved):
Thanks @MuhamedAbdalla! I made some tweaks, now merging, changes will be available in 1.8.7.

PR #2331, opened 2023-12-11
Fix: grammar error

twojnarowski:
Correct grammar fix, "stick" in 3rd form is stuck, not stucked.
https://ell.stackexchange.com/questions/85624/stucked-is-it-correct


odinserj:
Thanks @twojnarowski! 👍 👍 👍

PR #2350, opened 2024-01-17
Check if response `HasStarted` before changing it.

maliming:
ASP.NET Core does not buffer the http response body. This means that the very first time the response is written, the headers are sent along with that chunk of the body to the client. When this happens, it's no longer possible to change response headers.

https://github.com/davidfowl/AspNetCoreDiagnosticScenarios/blob/master/AspNetCoreGuidance.md#avoid-adding-headers-after-the-httpresponse-has-started

### DIFF
--- a/src/Hangfire.AspNetCore/Dashboard/AspNetCoreDashboardResponse.cs
+++ b/src/Hangfire.AspNetCore/Dashboard/AspNetCoreDashboardResponse.cs
@@ -35,25 +35,45 @@ namespace Hangfire.Dashboard
         public override string ContentType
         {
             get { return _context.Response.ContentType; }
-            set { _context.Response.ContentType = value; }
+            set
+            {
+                if (!_context.Response.HasStarted)
+                {
+                    _context.Response.ContentType = value;
+                }
+            }
         }
 
         public override int StatusCode
         {
             get { return _context.Response.StatusCode; }
-            set { _context.Response.StatusCode = value; }
+            set
+            {
+                if (!_context.Response.HasStarted)
+                {
+                    _context.Response.StatusCode = value;
+                }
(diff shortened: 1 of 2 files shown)


src/Hangfire.AspNetCore/Dashboard/AspNetCoreDashboardResponse.cs
```diff
@@ -35,25 +35,45 @@ public AspNetCoreDashboardResponse([NotNull] HttpContext context)
...
 
         public override Task WriteAsync(string text)
         {
-            return _context.Response.WriteAsync(text);
+            if (!_context.Response.HasStarted)
+            {
+                return _context.Response.WriteAsync(text);
+            }
+
+            return Task.CompletedTask;;
```
odinserj, line 68: Thanks a lot! Unfortunately some older .NET frameworks don't support `Task.CompletedTask`, can you change it to `Task.FromResult(true)` instead? By the way, do you get any exceptions due to the current behavior?

odinserj (changes requested):

src/Hangfire.AspNetCore/Dashboard/AspNetCoreDashboardResponse.cs
maliming: hi 

I changed it to `Task.FromResult(true)`. It was a customer feedback.

https://support.abp.io/QA/Questions/6514/StatusCode-cannot-be-set-because-the-response-has-already-started

odinserj: Thanks!

odinserj (approved):

PR #2353, opened 2024-01-22
add automatic retry action on retry exceeded

alexrosenfeld10:
I'm looking to provide a simple way to run a callback when retries are exceeded. This is useful for things like alerting when retries are fully exhausted. If there's a preexisting way to do this that I am not aware of, kindly point me towards it and close this PR. Thanks!

Disclaimer - I only have a mac and therefore could not compile and run this. Hoping the maintainers can take a peek and let me know what the next steps should be.


alexrosenfeld10:
@odinserj , wondering if you have any thoughts about adding a feature like this. Thanks in advance for your time

alexrosenfeld10:
@odinserj, would love to hear from you about this pr idea

PR #2363, opened 2024-02-08
Added Norwegian translations for new keys

khellang:
Added translations for new entries since the initial translation 😄


odinserj (approved):
Thanks a lot, merging!

PR #2377, opened 2024-03-06
2 minor localization errors in Strings.sv.resx

Uglack:
Found 2 very minor errors in the Swedish localization file.

I have not created an issue for this since it's such a small thing, but I can absolutely create one if it's needed! :)


odinserj (approved):
Thanks a lot for polishing! Delayed by signing issues, because secrets aren't available in PR builds, now everything succeeded.

PR #2393, opened 2024-04-16
Fix UrlHelper.To in Obsolete case

LordJZ:
The piece of code in question:
```C#
        public string To(string relativePath)
        {
            return _context.Options.PrefixPath +
                   (
#if FEATURE_OWIN
                       _owinContext?.Request.PathBase.Value ??
#endif
                       _context.Request.PathBase
                       + relativePath
                   );
        }

```

With FEATURE_OWIN this compiles to the following:
```C#
return _context.Options.PrefixPath +
    (_owinContext?.Request.PathBase.Value ??
        (_context.Request.PathBase + relativePath));
```

Note how in the Owin case the `relativePath` value is lost.

Not a big deal since the Owin case is deprecated via constructor attribute.

### DIFF
--- a/src/Hangfire.Core/Dashboard/UrlHelper.cs
+++ b/src/Hangfire.Core/Dashboard/UrlHelper.cs
@@ -48,8 +48,7 @@ namespace Hangfire.Dashboard
                        _owinContext?.Request.PathBase.Value ??
 #endif
                        _context.Request.PathBase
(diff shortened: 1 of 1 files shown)


odinserj (approved):

odinserj:
Thanks @LordJZ, definitely makes sense!

Issue #2408, opened 2024-05-24
Hangfire with Aspire integration


meghuizen:
What I think would be a killer combo is having Hangfire integration in Aspire.
The scenario I would like to see is that you have Aspire running with multiple Hangfire workers on different projects running in Aspire.
The Hangfire dashboard as plugin part of the Aspire dashboard, which will give an overview on all the Hangfire jobs on all the Hangfire workers running in Aspire.
Also the telemetry and logging being visible from Aspire.

odinserj:
Looks indeed interesting, but need to understand its component model first. It seems to me that it's possible to integrate Hangfire with its current capabilities, since Aspire is likely using ASP.NET Core infrastructure, so it is likely to see the logs in Aspire dashboard anyway. But more detailed integration would be useful to have.

adanbrownpaca:
We should make this simpler - we should add the ability for Hangfire to simply use a DB on the container provide by aspire. Do do this it needs to be able to install the DB seamlessly:

Hangfire Project:
builder.AddSqlServerClient("PacaBackgroundJobsDb"); 

Aspire:
var sqldbserver = builder.AddSqlServer("DbServer");
var backgroundjobsdb = sqldbserver.AddDatabase("BackgroundJobsDb");
builder.AddProject<Projects.BackgroundJobs>("backgroundjobs")
	.WithReference(backgroundjobsdb);

This way the Db gets installed on the local development container and initialized with the schema. Then, this simplifies deployment to Azure.

arielmoraes:
Let's not forget Aspire provides packages for the client too. So all the goodies like OpenTelemetry, Health Checks, and more are expected to be provided out-of-box.

sagarok:
> We should make this simpler - we should add the ability for Hangfire to simply use a DB on the container provide by aspire. Do do this it needs to be able to install the DB seamlessly:
> 
> Hangfire Project: builder.AddSqlServerClient("BackgroundJobsDb");
> 
> Aspire: var sqldbserver = builder.AddSqlServer("DbServer"); var backgroundjobsdb = sqldbserver.AddDatabase("BackgroundJobsDb"); builder.AddProject<Projects.BackgroundJobs>("backgroundjobs") .WithReference(backgroundjobsdb);
> 
> This way the Db gets installed on the local development container and initialized with the schema. Then, this simplifies deployment to Azure.

Would be very useful. Any chance on that?

sgryphon:
[...]

Issue #2017, opened 2022-03-27
Weird problem when running two methods inside the Enqueue


luizfbicalho:
I have an extension to enqueue my view models pointing to an implementation of an interface IBackgroundJob<T>

this are my extensions methods

```
private static readonly ActivitySource activitySource = new("MC.Hangfire.Extensions");

public static string Enqueue<T>(this T job, IBackgroundJobClient client)
{
    return client.Enqueue<IBackgroundJob<T>>(ps => ps.AddTelemetry(null).EnqueueJob(null, job, JobCancellationToken.Null));
}

public static IBackgroundJob<T> AddTelemetry<T>(this IBackgroundJob<T> job, PerformContext context)
{
    using var activity = activitySource.StartActivity($"Start Job {typeof(T).FullName} id {context.BackgroundJob.Id}", ActivityKind.Server);
    activity?.SetTag("JobId", context.BackgroundJob.Id);
    activity?.SetTag("JobJson", Newtonsoft.Json.JsonConvert.SerializeObject(job));
    activity?.SetTag("Job", Newtonsoft.Json.JsonConvert.SerializeObject(context.BackgroundJob.Job));
    return job;
}
```

My problem is that the EnqueueJob method is called, but the AddTelemetry method is not called before, how can I Add the telemetry information before calling all of my jobs, but in the context of the jobs, and of course not adding this code in all of my enqueue methods?

luizfbicalho:
I copied this filter and did this filter

```C#
public class TraceLogFilterAttribute : JobFilterAttribute,
IClientFilter, IServerFilter, IElectStateFilter, IApplyStateFilter
{
    public TraceLogFilterAttribute()
    {

    }

    private static readonly ILog Logger = LogProvider.GetCurrentClassLogger();

    public void OnCreating(CreatingContext context)
    {
        Logger.InfoFormat("Creating a job based on method `{0}`...", context.Job.Method.Name);
    }

    public void OnCreated(CreatedContext context)
    {
        Logger.InfoFormat(
            "Job that is based on method `{0}` has been created with id `{1}`",
            context.Job.Method.Name,
            context.BackgroundJob?.Id);
    }
    private static readonly ActivitySource activitySource = new("MC.Hangfire.Extensions");


    public void OnPerforming(PerformingContext context)
    {
[...]

PR #2460, opened 2024-10-27
Add ActivitySource to start producer and consumer activities (spans) to support OpenTelemetry observability

sgryphon:
OpenTelemetry is a cross-platform open-source standard for distributed tracing, which allows you to collect and analyze data about the performance of your systems.

OpenTelemetry is now the default for new .NET applications: https://learn.microsoft.com/en-us/dotnet/core/diagnostics/observability-with-otel

These changes add a Hangfire ActivitySource to start a Producer activity (span) when a job (either background or recurring) is created, and then a Consumer activity when it is processed.

It addresses part of the integration to Aspire, requested in https://github.com/HangfireIO/Hangfire/issues/2408

It also provides a built in core solution, compared to using filters discussed in https://github.com/HangfireIO/Hangfire/issues/2017

The creation context information is persisted with the job, so that the distributed activities can be correlated, even if processed on a different server.

No library user involvement or changes are needed except to register a listener for the ActivitySource, such as via OpenTelemetry. The Producer activity will pick up any existing context (such as from an ASP.NET request), or create a new one as needed.

The NetCore example has been updated to show how TraceId is correlated between job creation and execution. The PR also includes unit tests and a brief mention in the Readme.

Note: Some of the work has been based on the MassTransit implementation of ActivitySource.

### DIFF
--- a/NuGet.config
+++ b/NuGet.config
@@ -12,7 +12,7 @@
 
   <trustedSigners> 
     <repository name="nuget.org" serviceIndex="https://api.nuget.org/v3/index.json">
-      <owners>Microsoft;aspnet;dotnetframework;HangfireIO;xunit;jamesnk;kzu;castleproject;psake;ILRepack;davidebbo;StackExchange;Dapper;brady.holt;dwhelan;raboof;damianh;</owners>
+      <owners>Microsoft;aspnet;dotnetframework;HangfireIO;xunit;jamesnk;kzu;castleproject;psake;ILRepack;davidebbo;StackExchange;Dapper;brady.holt;dwhelan;raboof;damianh;OpenTelemetry;</owners>
       <certificate fingerprint="5a2901d6ada3d18260b9c6dfe2133c95d74b9eef6ae0e5dc334c8454d1477df4" hashAlgorithm="SHA256" allowUntrustedRoot="false" />
       <certificate fingerprint="0e5f38f57dc1bcc806d8494f4f90fbcedd988b46760709cbeec6f4219aa6157d" hashAlgorithm="SHA256" allowUntrustedRoot="false" />
       <certificate fingerprint="1f4b311d9acc115c8dc8018b5a49e00fce6da8e2855f9f014ca6f34570bc482d" hashAlgorithm="SHA256" allowUntrustedRoot="false" /> 
--- a/README.md
+++ b/README.md
@@ -110,6 +110,37 @@ using (new BackgroundJobServer())
 }
 ```
 
+Observability
+-------------
+
+**Open Telemetry**
+
+OpenTelemetry is an open-source standard for distributed tracing, which allows you to collect and analyze data about the performance of your systems. Hangfire can be configured to use OpenTelemetry to instrument message handling, so that you can collect telemetry data about messages as they flow through your system.
+
+Hangfire has a default filter that starts a Producer activity when a background job is scheduled, and start a Consumer activity when a job is performed. The distributed trace correlation information of the creation context is persisted with the job and used for the consumer (even if run on a different server), allows jobs to be correlated. Note that a recurring job creates a new activity each time it runs.
+
+For more information see: <https://opentelemetry.io/docs/specs/semconv/messaging/messaging-spans/>
+
+**Tracing**
+
+This example is using following packages:
+- `OpenTelemetry.Extensions.Hosting`
+- `OpenTelemetry.Exporter.Console`
+
+```csharp
+using OpenTelemetry.Trace;
+
+var builder = WebApplication.CreateBuilder(args);
+
+builder.Services.AddOpenTelemetry()
+    .WithTracing(b => b
+        .AddSource(DiagnosticsActivityFilter.DefaultListenerName) // Hangfire ActivitySource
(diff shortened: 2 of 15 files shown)


sgryphon:
@odinserj should I target this at the Main or Dev branch? Thanks.

odinserj:
Hi @sgryphon! I like the idea, but the number of changes is really high. How does it compare with the following gist that's implemented as a job filter? https://gist.github.com/odinserj/06e18950b5bf3083a5aed0ed06d3d18a

sgryphon:
> How does it compare with the following gist that's implemented as a job filter? 

Thanks for reviewing.

I am currently use a similar (but limited) filter, as discussed in https://github.com/HangfireIO/Hangfire/issues/2017

I do think that activity tracing, like metrics, is better as part of the core solution, turned on by default / with a standard name.  I based some of that on MassTransit, where you just subscribe to a known event source; the .NET elements (e.g. HttpClient) are similar in that you can just easily enable them.

Decoupling from the flow is however a good idea, so happy with a different approach. My main thoughts are:

1. Can the filter be a default system filter, i.e. turn on by default (with a fixed Activity Source name).
2. Does it support both background jobs and recurring jobs.

Apart from that, some specific feedback on the gist:

* OpenTelemetry has some specific conventions around span naming, e.g. operation.name is usually a single token (i.e. 'perform_job' instead of 'perform job'). See https://opentelemetry.io/docs/specs/semconv/messaging/messaging-spans/#span-name
* While you can add custom tags, like 'job.state' or 'job.storage', there are semantic conventions for many common tags, e.g. 'messaging.destination.name' should match the second part of the span name, 'message.operation.type' should be from a specific list, etc.
* I don't think you need to store created at, because spans are timestamped anyway.
* Not sure why OnCreating has two StartActivity; the Consumer span should be a child of the Producer span, i.e. only one span needed (that is passed through as the parent to the second)
* Should pass through both activity.Id and activity.TraceStateString (even though we aren't using trace state, an upstream trace may). 
* For consistency (e.g. with HTTP) the job parameter names should be "traceparent" and "tracestate" (HTTP is not case sensitive, so often as "TraceParent")
* Not sure if isRemote needs to be set (if it is, maybe check it is remote??)
* Just double check that activity creates correctly is parent is null/empty (e.g. from a client that doesn't set it). I think older versions of the library didn't handle empty/null well.
* SetStatus(OK) ignores the description string (so not needed).
* SetStatus(Error) is only in newer (6+) library; otherwise set attributes separately (in my code). Fine if you are using the newer library. There is also an AddException() in .NET 9.

If you can answer the first two items, I'd be happy to work on a revised solution using the filter.

sgryphon:
> 1. Can the filter be a default system filter, i.e. turn on by default (with a fixed Activity Source name).
> 2. Does it support both background jobs and recurring jobs.

I found `GlobalJobFilters.cs`, which I think answers my first question.

I will work on using the alternative approach. It is nicer, as it neatly decouples the work. I also like putting the activity in the context (doesn't work for job parameters, which may be serialized to another machine, but I presume the start and end events run in the same process).

sgryphon:
I've re-written the code as a global filter. I think it is much nicer to decouple like that -- and is shows the good architecture/structure of Hangfire that it is easy to use.

I used your filter code as a base, and then merged with my code (& applied my suggestions).

I added it as default to global filters, and then checked that the samples run okay (updated readme & samples for new location of the constant).

Some notes:

* I didn't use the IApplyStateFilter. It seems largely separate to the job processing. (One of the main goals was to get a TraceID for the performed job).
* Answering my question no. 2, it doesn't correlate recurring jobs. Although I was kind of on the fence about that anyway, i.e. each separate run of a recurring job gets its own TraceID (which meets the goal of having a TraceID when performing). In some respects this is better than the same TraceID for all recurring jobs.

The one thing I haven't updated yet is the unit tests, as the previous code was built into Worker, etc so I had tests for those. I'll have a look at the tests for other filters and see if I can add anything valuable.

sgryphon:
New unit tests, for the filter, now added.

odinserj:
Thanks @sgryphon, looks really good! One caveat here is related to a new dependency included in Hangfire.Core, and it's version is deprecated as per NuGet. Given .NET 9.0 became strict against transitive dependencies (as in https://github.com/HangfireIO/Hangfire/issues/2468), I'm thinking on the following steps:

1. Bring `net6.0` platform to Hangfire.Core in the `dev` branch.
2. Rebase this PR onto the new branch.
3. Enable OpenTelemetry support only on `net6.0` platform with the corresponding version of `System.Diagnostics.DiagnosticSource` that's not deprecated.

So on the modern platforms we'll have out-of-box support for telemetry without deprecation-related problems, for other platforms it will be still possible to use https://www.nuget.org/packages/OpenTelemetry.Instrumentation.Hangfire. 

What do you think?

sgryphon:
> What do you think?

Yes. I did see this message and agree, just haven't had time to do it yet :-)

willianantunes:
Hi all! Is this PR still on the roadmap for a merge soon? It would be great. 😁

odinserj:
Yes, it is on the roadmap for Hangfire 1.9.0, but this version requires some more time than expected.

PR #2574, opened 2026-07-13
Add a Russian translation for a Hangfire Dashboard

akortunov:
Implements #2548.

Note that some texts in the selected task's details are not covered by resource system-based translations at all. 

Also date and time texts (like "a minute ago") respected current locale with Hangfire 1.7.x, but do not in upstream code. This issue presumably is caused by commit 061999aacd226faf253b3ee91f8541cdee28af2e. To solve it, I need to return a Russian locale definition to the custom `moment-with-locales.min.js`. It will be done in the follow-up PR.

<details>
<summary>Some screenshots</summary>

image

image

image

image
</details>

PR #2576, opened 2026-07-15
Improve a support for locales with region subtags

akortunov:
A follow-up work for #2574.

In upstream Hangfire, locale name is passed to Hangfire Dashboard HTML as a two-letter language name:
```
<html lang="@CultureInfo.CurrentUICulture.TwoLetterISOLanguageName">
```

This tag is used as a locale name for Moment.js:

image

As a result, some supported locales become unused in Moment.js:
1. `pt-br` becomes just  `pt` instead
2. `zh-cn` and `zh-tw` become just `zh`, which does not exist in Moment.js, so an English text is used instead

A proposed solution:
1. Add a new attibute to `Hangfire.config`, which will store the full locale name.
2. Try to set a locale for Moment.js by using a full name first, and fall back to current behaviour if target locale was not found.

### DIFF
--- a/src/Hangfire.Core/Dashboard/Content/js/hangfire.js
+++ b/src/Hangfire.Core/Dashboard/Content/js/hangfire.js
@@ -49,7 +49,8 @@
     hangfire.config = {
         pollInterval: $("#hangfireConfig").data("pollinterval"),
         pollUrl: $("#hangfireConfig").data("pollurl"),
-        locale: document.documentElement.lang,
+        lang: document.documentElement.lang,
+        locale: $("#hangfireConfig").data("locale"),
         darkMode: $("#hangfireConfig").data("darkmode")
     };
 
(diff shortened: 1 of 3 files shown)


odinserj:
Thanks @akortunov! I will check this a bit later, sorry for the delay – need also to dig into the `lang` attribute to understand whether locale can be used there, not just the two-letter codes.

burningice2866:
> Thanks @akortunov! I will check this a bit later, sorry for the delay – need also to dig into the `lang` attribute to understand whether locale can be used there, not just the two-letter codes.

Yeah, its allowed to use the full lang-region value in the html-attribute

https://developer.mozilla.org/en-US/docs/Glossary/BCP_47_language_tag

akortunov:
> > Thanks @akortunov! I will check this a bit later, sorry for the delay – need also to dig into the `lang` attribute to understand whether locale can be used there, not just the two-letter codes.
> 
> Yeah, its allowed to use the full lang-region value in the html-attribute
> 
> https://developer.mozilla.org/en-US/docs/Glossary/BCP_47_language_tag

Yes, it is possible (I suppose that it would be enough to replace `@CultureInfo.CurrentUICulture.TwoLetterISOLanguageName` to `@CultureInfo.CurrentUICulture.Name`).

But this change will affect the whole page instead of just moment.js and may have side effects. 

For example - I suppose that `ca-ES` locale will fall back to `en` instead of `ca` in `moment.js`.

PR #2580, opened 2026-07-17
Fix grammar and spelling in user-facing messages

net0well:
Correct several long-standing typos in messages that are surfaced to users through logs, exceptions and the Dashboard UI:

- AutomaticRetryAttribute: "an exception occured" -> "an exception occurred" in the warning logged when a job is automatically deleted after exceeding the maximum number of retry attempts.

- CoreBackgroundJobFactory: "because it doesn't exists" -> "because it doesn't exist" in the InvalidOperationException thrown when a background job can't be initialized during creation retries.

- DisableConcurrentExecutionAttribute: "greater that zero" -> "greater than zero" in the ArgumentException thrown when a negative timeout value is passed to the constructor.

- Dashboard (Strings.resx): "Cron expression is invalid or don't have any occurrences" -> "Cron expression is invalid or doesn't have any occurrences" in the disabled recurring job resource files are not affected since each culture provides it  value for this key.

Only string literals are changed, so there is no behavior change, and no existing tests assert these exact messages.

### DIFF
--- a/src/Hangfire.Core/AutomaticRetryAttribute.cs
+++ b/src/Hangfire.Core/AutomaticRetryAttribute.cs
@@ -378,7 +378,7 @@ namespace Hangfire
             if (LogEvents)
             {
                 _logger.WarnException(
-                    $"Failed to process the job '{context.BackgroundJob.Id}': an exception occured. Job was automatically deleted because the retry attempt count exceeded {Attempts}.",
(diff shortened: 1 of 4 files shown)


odinserj (approved):

odinserj:
Thank you, @net0well!

PR #2581, opened 2026-07-20
Add unit tests for DisableConcurrentExecutionAttribute

net0well:
Cover constructor validation, distributed lock acquisition with both default and formatted resource identifiers, and lock release behavior.


odinserj (approved):

odinserj:
Thanks @net0well!

PR #2589, opened 2026-08-24
Don't disable recurring jobs on transient storage errors

### LINKED ISSUE #2582: Recurring jobs get permanently disabled when the database is full (SQL error 1105)
On SQL Server Express, which has a 10 GB limit per database. When it filled up, all our recurring jobs quietly stopped and never came back — even after we freed space. We had to restart the app to get them running again.

Digging in: when a recurring job can't be scheduled because the DB is full, job creation throws `SqlException` 1105 ("filegroup is full"). `RecurringJobScheduler` retries 5 times (~5 min) and then disables the job — `NextExecution = null`, score `-1` in the `recurring-jobs` set. Since the scheduler never looks at negative scores again, the job is dead until something calls `AddOrUpdate` (app restart) or you trigger it by hand.

That makes sense for a genuinely broken job (bad cron, unloadable type), but a full disk or a timeout is a temporary storage problem, not a broken job — so disabling it permanently feels wrong.

**Repro:** cap a SQL Server data file (`MODIFY FILE (MAXSIZE = 72MB, FILEGROWTH = 0)`), fill it with junk until inserts fail with 1105, then run a `* * * * *` recurring job. Within ~5 minutes it's disabled, and freeing the space afterwards doesn't bring it back.

**Suggestion:** in `RetryRecurringJob`, don't count storage errors toward the disable limit — just keep retrying so the job recovers on its own once the DB is healthy. Checking for `DbException` would cover timeouts and connection drops too, but even just handling 1105 would fix the common case. Real job-definition errors would still disable as they do today.

Happy to open a PR if you're open to it.

HarnageaGabriel:
## Summary

Fixes #2582. `RecurringJobScheduler.RetryRecurringJob` decided retry vs. permanent disable using retry-attempt count alone, with no regard to the exception's cause. A transient storage failure (e.g. SQL Server error 1105 "could not allocate space due to full filegroup", or any dropped connection/timeout) was therefore treated the same as a genuinely broken job (bad cron expression, unloadable type) and eventually disabled the job permanently, requiring manual re-enable — exactly as described by the original report and confirmed via source-level analysis in the issue comments (thanks @akarboush and @arnelirobles).

Per the options weighed in the issue thread, this implements the least invasive of the three: classify exceptions whose chain contains a `System.Data.Common.DbException` as transient storage errors, in `Hangfire.Core` itself (no public API addition, no storage-package coupling). Such failures keep retrying with the existing delay and are not counted toward the disable threshold. Non-`DbException` failures keep the prior behavior — retry up to `MaxRetryAttemptCount`, then disable.

## Changes

- `src/Hangfire.Core/Server/RecurringJobScheduler.cs`: added `IsTransientStorageException`, walks the exception's `InnerException` chain for a `DbException`. Transient failures skip the disable branch and don't increment the retry-attempt counter.
- `src/Hangfire.Core/RecurringJobEntity.cs`: `ScheduleRetry` gained an optional `incrementRetryAttempt` parameter (default `true`, preserving existing behavior) so transient retries don't consume the attempt budget.
- `tests/Hangfire.Core.Tests/Server/RecurringJobSchedulerFacts.cs`: new theory covering direct and wrapped `DbException`, both batching modes, retry count exceeding the old threshold, confirming the job is rescheduled (not disabled) and its score never becomes `-1`.

## Test plan

- [x] `dotnet build` — 0 warnings, 0 errors
- [x] `dotnet test` on `Hangfire.Core.Tests`, net8.0: 1,520 passed, 0 failed
- [x] `RecurringJobSchedulerFacts`: 107 passed, 0 failed (incl. 4 new)
- [x] net452: 1,470 passed; net461: 1,508 passed

### DIFF
--- a/src/Hangfire.Core/JobStorage.cs
+++ b/src/Hangfire.Core/JobStorage.cs
@@ -127,7 +127,7 @@ namespace Hangfire
         /// storage-level failure (connection drop, timeout, deadlock victim, failover, etc.)
         /// that is expected to succeed on a later retry. Storage implementations should
         /// override this to recognise their own transient failure conditions, since a
-        /// generic exception type alone (for example <see cref="System.Data.Common.DbException"/>)
+        /// generic exception type alone (for example <c>System.Data.Common.DbException</c>)
         /// is not enough to tell transient and permanent errors apart.
         /// </summary>
         /// <param name="exception">The exception to inspect. May be <see langword="null"/>.</param>
--- a/src/Hangfire.Core/JobStorage.cs
+++ b/src/Hangfire.Core/JobStorage.cs
@@ -121,5 +121,20 @@ namespace Hangfire
             if (featureId == null) throw new ArgumentNullException(nameof(featureId));
             return false;
         }
+
+        /// <summary>
+        /// Determines whether the given exception is a transient one, i.e. a temporary
+        /// storage-level failure (connection drop, timeout, deadlock victim, failover, etc.)
+        /// that is expected to succeed on a later retry. Storage implementations should
+        /// override this to recognise their own transient failure conditions, since a
+        /// generic exception type alone (for example <see cref="System.Data.Common.DbException"/>)
+        /// is not enough to tell transient and permanent errors apart.
+        /// </summary>
+        /// <param name="exception">The exception to inspect. May be <see langword="null"/>.</param>
+        /// <returns><see langword="true"/> if the exception is considered transient; otherwise <see langword="false"/>.</returns>
+        public virtual bool IsTransientException([CanBeNull] Exception exception)
+        {
+            return false;
+        }
     }
 }
--- a/src/Hangfire.Core/RecurringJobEntity.cs
+++ b/src/Hangfire.Core/RecurringJobEntity.cs
@@ -186,9 +186,13 @@ namespace Hangfire
             return changedFields.Count > 0;
         }
 
-        public void ScheduleRetry(DateTime nextAttempt, string error)
+        public void ScheduleRetry(DateTime nextAttempt, string error, bool incrementRetryAttempt = true)
         {
-            RetryAttempt++;
+            if (incrementRetryAttempt)
+            {
+                RetryAttempt++;
+            }
+
             Error = error;
             NextExecution = nextAttempt;
         }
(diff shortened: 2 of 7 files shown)


odinserj:
Hi @HarnageaGabriel! Thanks for the PR – I like the idea of introduction of transient storage exceptions, because it is indeed useful for dealing with some exceptions. 

However, I think we should delegate the logic of determining whether a particular exception is transient to the Storage API, for example by introducing a virtual `IsTransientException(Exception) method in the `JobStorage` class. In this case, the implementation will be more flexible and would not depend only on ADO.NET. Additionally, it will be possible to use storage specifics (like SQL Server's one), because not every `DbException` is a transient one.

HarnageaGabriel:
Thanks for the review, @odinserj! I've reworked the PR along the lines you suggested.

- Added `virtual bool IsTransientException(Exception exception)` to `JobStorage`. The base implementation is deliberately conservative and returns `false`, so it carries no ADO.NET dependency and makes no assumptions on behalf of a storage.
- `RecurringJobScheduler.RetryRecurringJob` now calls `storage.IsTransientException(error)` (the storage is threaded in from `context.Storage`) instead of walking the exception chain for `DbException` itself. The `System.Data.Common` dependency is gone from the scheduler.
- `SqlServerStorage` overrides `IsTransientException` with the inner-exception `DbException` walk that previously lived in the scheduler. Kept simple for now; narrowing it further with SQL Server specifics (e.g. `SqlException.Number` for the well-known transient error codes) is an easy follow-up if you'd like it in this PR.
- Tests updated to the new design: `JobStorageFacts` covers the conservative default, `RecurringJobSchedulerFacts` now drives the behaviour through a mocked `IsTransientException` (transient -> reschedule without incrementing `RetryAttempt`; non-transient -> disable once attempts are exhausted), and `SqlServerStorageFacts` covers the override for direct/wrapped/non-`DbException`/null.

`dotnet test` for the `RecurringJobScheduler`/`JobStorageFacts` filter (115 tests) and the `SqlServerStorageFacts.IsTransientException` filter (4 tests) pass locally on net8.0.

akarboush:
Opened #2591, which hits the same class of problem in Worker/DelayedJobScheduler: a transient storage failure gets recorded as a permanent Failed state with disableFilters: true, which bypasses the retry policy and orphans every continuation. It needs the same transient-exception classification you're being asked to move behind the Storage API here, so that hook would have a second consumer.

